PR #4, opened 2021-06-16
Add helm chart testing

macobo:
- Make chart k3s friendly (for CI)
- Add testing for the helm chart
    
The following tests are currently made:
1. It installs the chart on various k8s versions
2. Verifies whether the chart comes up cleanly (all pods eventually start)
3. Sends a bunch of events via k6 test script
4. Verifies they are all ingested (via trends API)
5. Tests whether one can upgrade cleanly from a chart in the same state
as `main` branch

In addition the action:
- Lists information on the kubernetes cluster after every test run
- Displays a kubernetes manifest diff from main branch chart under the upgrade test

Recommend taking a peek at the output, we'll be looking at it in the future!

Follow-up PRs will be made for further testing methods + linting, trying to keep the PR lean.

More context (including charts this draws inspiration from) in https://github.com/PostHog/vpc/issues/58

Note that upgrade tests are expected to fail on this branch - main branch does not deploy cleanly on CI.

### DIFF
--- a/.github/workflows/test-helm-chart.yaml
+++ b/.github/workflows/test-helm-chart.yaml
@@ -106,7 +106,7 @@ jobs:
 
       # GitHub Action reference: https://github.com/k6io/action
       - name: Run ingestion test using k6
-        uses: k6io/action@v0.1
+        uses: k6io/action@v0.2.0
         with:
           filename: ci/k6-ingestion-test.js
 


.github/workflows/test-helm-chart.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+      - name: Setup for ingestion test
+        run: |
+          export WEB_POD=$(kubectl get pods -l role=web -o jsonpath="{.items[].metadata.name}")
+          kubectl exec $WEB_POD -- python manage.py setup_dev --no-data
+
+          echo "POSTHOG_HOSTNAME=$(kubectl get svc posthog -o jsonpath="{.spec.clusterIP}")" >> $GITHUB_ENV
+
+          # Sleep until web service is ready to respond
+          sleep 20
```
tiina303, line 105: could we use the pod status (same as load balancer uses to know if to send traffic)?

macobo: The above actions wait for the deployment to be ready, which should be the same thing but in practice the first few seconds of requests always failed due to address propagation or smth. Since we don't have a reliable way to wait for that - this. Seems to work great in practice.

PR #5, opened 2021-06-17
Update README.md, add warning + fix intro

macobo:


charts/posthog/README.md
```diff
@@ -1,34 +1,37 @@
...
-$ helm install --timeout 20m posthog .
-```
+🦔 [PostHog](https://posthog.com/) is developer-friendly, open-source product analytics.
 
 ## Introduction
 
 This chart bootstraps a [PostHog](https://posthog.com/) deployment on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
 
-This installation is intended for larger-scale usage (> 1M events/month) an
+This installation is geared towards larger scalability (> 1M events per month).
```
tiina303, line 11: I thought the plan was everyone uses this, so I'd just drop this line completely?

macobo: Please add a note to the EPIC ticket or drop the PR you'd like. I like it here since it helps justify why this exists vs the main chart.

PR #7, opened 2021-06-17
Remove some dead files

macobo:
@fuziontech if you think the files are needed, let me know under this
PR.


fuziontech (approved):
Solid cleanups ✅

PR #14, opened 2021-06-17
Improve defaults, make chart more small-installation friendly

macobo:
We still need to write good docs (e.g. how to get started) but these are
I hope saner defaults to start from.


fuziontech (approved):
perfect

PR #15, opened 2021-06-18
Values files rfc

tiina303:
Three things:
#### 1. bump down resource requests (storage)
Please advise on what are actually reasonable values here

#### 2. minimal values files

We have a lot of values that users can override & we likely want to keep it that way. However we also want to make it easy for them to get started, so the proposal here is to provide the values files that contain minimal values users need to set (that they can supply with `-f` flag to `helm` later).

Do we need email to be set up?

Secondly there are different scales at which users need to ask for resources differently. Please advise on the values/scales, but also feedback on the approach of providing different values files like this.

#### 3. hpa to theoretically work

I wanted to make the third scale up to whatever file contain autoscaling even though we haven't really tested it. Because we already had some barebones in `values.yaml` and that could have been a bit confusing otherwise.

My understanding of how things work:
1. if hpa is enabled it changes the deployment replicas value ([source](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/)) 
2. In order for hpa to work reasonably resource limits should be set (for cpu that we use to scale by) ([source](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/)) 
> Note: Since the resource usages of all the containers are summed up the total pod utilization may not accurately represent the individual container resource usage. This could lead to situations where a single container might be running with high usage and the HPA will not scale out because the overall pod usage is still within acceptable limits.

3. There was a [problem](https://github.com/kubernetes/kubernetes/issues/25238) in how hpa worked during upgrades. It was solved in 1.19 & we should leave replicas empty if hpa is used.

Note that I added some more copy-pasta & I'd love to clean that up at one point to make it easier to read and more typo safe, but it's not a priority atm.

## Test plan
no errors from:
```
helm install --debug --dry-run posthog charts/posthog -f values.yaml
helm install --debug --dry-run posthog charts/posthog -f values-10k-to-1M.yaml
helm install --debug --dry-run posthog charts/posthog -f values-more-than-1M.yaml
```


macobo:
> (that they can supply with -f flag to helm later).

The problem here is that a lot of these resources are _not_ easily resizable after the fact. 

Because of that I suggest:

1. Users should instead specify proper values upfront - our README should contain a proper 
1. Because of that these values seem way too low - users will run into issues way too soon. 

I recommend trying out deploying a new small test cluster on GKE and seeing how it operates under real load - e.g. adapting my k6 test and sending ~10M events to it, seeing whether autoscaler works etc.

charts/posthog/values.yaml
```diff
@@ -251,7 +266,7 @@ ingress:
   hostname:
   gcp:
     # -- Specifies the name of the global IP address resource to be associated with the google clb
-    ip_name:
+    ip_name: "posthog"
```
macobo, line 269: Why? Removed it on purpose.

values-10k-to-1M.yaml
```diff
@@ -0,0 +1,32 @@
+# Minimal settings for scale of 10k to 1M events/month.
```
macobo, line 1: What's the purpose of these files? As-is users will not look here.

I suggest writing a proper README instead (or another doc and linking it from README), outlining the configuration options etc.

charts/posthog/templates/beat-hpa.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.beat.hpa.enabled -}}
```
macobo, line 1: Why does celery beat need autoscaling in the first place?

charts/posthog/values.yaml
```diff
@@ -13,7 +13,7 @@ image:
   # -- Image pull policy
   pullPolicy: IfNotPresent
 
-# -- Cloud service being deployed on. Either `gcp` or `aws`
+# -- Cloud service being deployed on. Only `gcp` is supported at the moment.
```
macobo, line 16: That's not really accurate. We do support `aws` in the chart - but we won't have setup instructions, etc on here. 

Let's keep this as-is. README.md is a better medium for explaining supported clouds.

charts/posthog/templates/plugins-deployment.yaml
```diff
@@ -13,7 +13,9 @@ spec:
         app: {{ template "posthog.fullname" . }}
         release: "{{ .Release.Name }}"
         role: plugins
+  {{- if not .Values.plugins.hpa.enabled }}
```
macobo, line 16: Have you tried out the autoscaling on plugins?

I don't think the hpa will work as-is on it - from what I've seen in the past it's not really cpu-bound even when it's falling behind.

charts/posthog/templates/pgbouncer-hpa.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.pgbouncer.hpa.enabled -}}
```
macobo, line 1: Will this work? Postgres has pretty strict connection limits internally and each pgbouncer takes up N connections - scaling pgbouncers out of bounds can easily lead to running out of connections -> errors & outages.

macobo:
Q: What is "test plan"?

charts/posthog/values.yaml
tiina303: Because we get an error if this isn't specified and I wasn't sure if that needs to be something our users need to specify or if it's fine for this to be `"posthog"`. I could also try to look into the failure we see from `helm install --debug --dry-run posthog charts/posthog`, but this seemed easier.

macobo: Issue being that while install will work with this specified, it will not work - the instance will not be accessible unless there actually is an IP with that name created. 

See https://posthog.com/docs/deployment/deploy-gke-clickhouse#step-by-step-installation

values-10k-to-1M.yaml
tiina303: I haven't updated the readme, but I was planning to do that. The purpose is to make it easier and faster/easier/less error prone for the users for the initial install, will help increase engagement/activation and reduce support load. 

To elaborate the thinking was: as a user I have no idea how much resources various parts in Posthog need, but I do know how many events / month I get. If based on the latter we direct them to reasonably working config that would be great. 

In the beginning we're not going to be good (well I'm not yet Karl and James might be) at suggesting what the config should be, but over time with more testing and with the feedback from the community we'll get to a good config and then it's possible to get to very small install times. 

If we have a large values file with all possible configs, it's much harder for a new person to know what they need to be doing hence it will take more time (we can use comments like "you must set this" etc, but I think it's easier for our users with these short files). Files rather than only text/explanation in a readme is another attempt to reduce time cost for the initial install. 

I might be wrong about all of this, but this is what my intuition is telling me.

charts/posthog/templates/beat-hpa.yaml
tiina303: Will remove. I wasn't sure if this could run into scaling problems at one point if we have small pods, but seems like multiple instances aren't necessarily safe to run which takes priority.

charts/posthog/templates/plugins-deployment.yaml
tiina303: Nope, didn't try it, only changed some parts about it (because I was confused about what I saw across various values files, so I did some reading). I'd love to look into this more, but I suspect it's not a priority atm. The alternative I considered was removing everything tied to autoscaling from this repo, but given that we likely want to add something about it in the future I opted not to.

tiina303:
> Q: What is "test plan"?

Something we used to always provide at FB, I really liked having that on diffs as it made it really easy for me to figure out what if any tests to run if I changed something I or others had worked on before.

charts/posthog/values.yaml
tiina303: discussed offline - keeping this, will make the documentation be clear that the name must match with the IP name they set (& that the default is posthog)

values-10k-to-1M.yaml
```diff
@@ -0,0 +1,32 @@
...
+  master:
+    size: 10Gi
+
+kafka:
+  persistence:
+    size: 10Gi
+
+# Extra replica for more loaded services
+pgbouncer:
+  replicacount: 2
```
macobo, line 26: Note - small installations will not need pgbouncer replicas, it can scale to enough scale.

values-10k-to-1M.yaml
```diff
@@ -0,0 +1,32 @@
+# Minimal settings for scale of 10k to 1M events/month.
+# Note that this is experimental, please let us know how this worked for you.
+
+ingress:
+  # -- URL to address your PostHog installation. You will need to set up DNS after installation. GCP tutorial: https://cloud.google.com/dns/docs/tutorials/create-domain-tutorial
```
macobo, line 5: I'd avoid linking to 3rd order documentation here. We already have:
- Docs in posthog.com
- README.md (incoming)
- Notes.txt

We should instead double down on having proper guides in 2 of the three.

values-10k-to-1M.yaml
```diff
@@ -0,0 +1,32 @@
...
+  hostname: "posthog.yourcloud.net"
+
+# x2 storage
```
macobo, line 8: This comment is confusing as-is, removing in my PR.

tiina303:
Rethinking the naming here: probably better to name all the `values.yaml` files (i.e. this one could be `values-small.yaml`) to make sure people think about that choice. Otherwise they will likely use this one accidentally ... it not name sharing with all values file helps avoid confusion too. If we at all will provide them in the repo

macobo:
Note - the values files now exist in readme. Let's avoid duplicating them :)

PR #26, opened 2021-06-29
Make upgrading with custom ingress.gcp.ip_name work

macobo:
The go template had a bug where it removed all preceding whitespace
before static-ip-name, resulting in invalid yaml.

Also used a more standard way to check for if value exists.

### DIFF
--- a/charts/posthog/templates/_helpers.tpl
+++ b/charts/posthog/templates/_helpers.tpl
@@ -331,7 +331,7 @@ Create the name of the service account to use
 {{- end -}}
 
 {{- define "ingress.type" -}}
-{{- if ne (.Values.ingress.type | toString) "<nil>" -}}
+{{- if hasKey .Values.ingress "type" -}}
   {{ .Values.ingress.type }}
 {{- else if .Values.ingress.nginx.enabled -}}
   nginx
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -13,7 +13,7 @@ metadata:
     kubernetes.io/ingress.class: "gce"
     networking.gke.io/managed-certificates: "{{ .Release.Name }}-gke-cert"
     networking.gke.io/v1beta1.FrontendConfig: "{{ .Release.Name }}-frontend-config"
-    {{- if .Values.ingress.gcp.ip_name -}}
+    {{- if .Values.ingress.gcp.ip_name }}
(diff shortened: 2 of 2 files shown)


charts/posthog/templates/_helpers.tpl
```diff
@@ -331,7 +331,7 @@ Create the name of the service account to use
 {{- end -}}
 
 {{- define "ingress.type" -}}
-{{- if ne (.Values.ingress.type | toString) "<nil>" -}}
+{{- if hasKey .Values.ingress "type" -}}
   {{ .Values.ingress.type }}
 {{- else if .Values.ingress.nginx.enabled -}}
   nginx
```
tiina303, line 337: That's what I tried initially, but sadly it doesn't work like I wanted directly and perhaps there is a better way or I should have left a comment explaining it.
Current world:
1. Our default values file has `ingress.type` and it's empty
2. The goal here was if the ingress type is specified use that, otherwise use nginx iff it's enabled, clb for gcp cloud and "undefined" otherwise (maybe erroring out for that would be better).

If ingress.type isn't specified we currently get true for hasKey, because it's defined, it's just empty.
```
> helm template posthog charts/posthog | grep -A1 HELM_INSTALL_INFO
        - name: HELM_INSTALL_INFO
          value: "{\"chart_version\":\"1.4.34\",\"cloud\":\"gcp\",\"hostname\":null,\"ingress_type\":\"\\u003cno value\\u003e\",\"operation\":\"install\",\"release_name\":\"posthog\",\"release_revision\":1}"
```

There are two ways we can make the goal (2) above. 
1. use this more complicated check that I had before that checks against non-empty (I think this is better)
2. remove `ingress.type` from `values.yaml`, but then users would need to know that the key they need to define is `ingress.type` plus our readme (ALL_VALUES file) wouldn't document it as an option then.

Q: Is ingress something we want to force that needs to be defined always, if yes I propose we error out in the else case, where I atm just say "undefined" otherwise I suggest adding a comment explaining why this if is more complex.

macobo: Q: Is ingress something we want to force that needs to be defined always, if yes I propose we error out in the else case, where I atm just say "undefined" otherwise I suggest adding a comment explaining why this if is more complex.

Yes - for posthog to be useful it needs to interact with the outside world. If we haven't tested a setup, we should assume it doesn't work and making it fail explicitly is useful. Feel free to go ahead and create a PR.

PR #29, opened 2021-06-30
Fix migrations and make TLS optional based on redirectToTLS and letsencrypt values

fuziontech:
Deployments to DigitalOcean were hanging and throwing a few errors blocking progress on getting this into the DigitalOcean Marketplace

### DIFF
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -27,7 +27,7 @@ metadata:
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/healthcheck-path: "/_health"
    {{- end }}
-   {{- if (eq (include "ingress.type" .) "nginx") }}
+   {{- if (and (eq (include "ingress.type" .) "nginx") (eq (include "ingress.redirectToTLS" .) "true" )) }}
     nginx.ingress.kubernetes.io/ssl-redirect: "true"
     nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
    {{- if eq (include "ingress.letsencrypt" .) "true"}}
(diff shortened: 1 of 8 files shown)


charts/posthog/templates/web-deployment.yaml
```diff
@@ -7,6 +7,12 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  annotations:
```
macobo, line 10: Confused - why does the web deployment + workers need this exactly?

charts/posthog/templates/ingress.yaml
```diff
@@ -40,15 +40,19 @@ metadata:
...
+  {{- if eq (include "ingress.letsencrypt" .) "true"}}
   tls:
   - hosts:
     - {{ .Values.ingress.hostname }}
     secretName: nginx-letsencrypt-{{ template "posthog.fullname" . }}
   {{- end }}
   rules:
-    - host: {{ .Values.ingress.hostname }}
+  {{- if .Values.ingress.hostname }} 
+    - host: {{ .Values.ingress.hostname | default "*" | quote }}
```
macobo, line 51: Does the default do anything here?

charts/posthog/templates/web-deployment.yaml
fuziontech: The migrations never were triggered on DigitalOcean. It could have been related to the `atomic` flag on helm. Basically with the `post-install` hook was waiting for `web` and `plugin` deployments to start, which wouldn't because the migrations never ran. This stages it so that as soon as the rest of the stack is ready: Run migrations, and then start plugins and web deployments. In that order.

charts/posthog/templates/ingress.yaml
fuziontech: This can definitely be cleaned up

Issue #27, opened 2021-06-29
Upgrade redis subchart


macobo:
We're currently 6 major versions behind upstream chart https://github.com/bitnami/charts/tree/master/bitnami/redis. Before we make the chart public, it might be worthwhile to get up to latest.

Feature(s) which are worthwhile for us: turning off replicas for small installations.

Upgrading for existing users: It seems like deleting statefulsets before upgrading might be needed.

PR #31, opened 2021-07-05
Update redis chart version

macobo:
Closes https://github.com/PostHog/charts-clickhouse/issues/27

The new version allows running redis in standalone mode - using 1 pod instead of 3. Given that redis is an ephemeral service for us, that seems like a win.

This is a breaking change - hence bumping the major version of the chart + documentation.

Tested on ee-test, the upgrade instructions are accurate. :)


charts/posthog/README.md
```diff
@@ -288,3 +290,15 @@ Simply apply one of the following labels to your PR _before merging_ to bump the
...
+
+### Upgrading from 1.x.x
+
+2.0.0 updated redis chart in an incompatible way. If your upgrade fails with
+
+```
+Upgrade "posthog" failed: cannot patch "posthog-posthog-redis-master" with kind StatefulSet: StatefulSet.apps "posthog-posthog-redis-master" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
+```
+
+Run the following and `helm upgrade` again: `kubectl delete sts posthog-posthog-redis-master`
```
fuziontech, line 304: does `kubectl delete sts posthog-posthog-redis-master` need a namespace to run? I think it does (just tried it on test DO cluster)

charts/posthog/values.yaml
```diff
@@ -303,16 +303,18 @@ redis:
...
-  # host: redis
-  # Just omit the password field if your redis cluster doesn't use password
-  # password: redis
-  # port: 6379
-  # When defined the `password` field is ignored
-  # existingSecret: secret-name
-  # existingSecretKey: redis-password
+
+  # -- Either standalone or cluster.
+  architecture: standalone
```
fuziontech, line 308: nice

fuziontech (approved):
Just need to add namespace to the delete kubectl command and this looks good

PR #37, opened 2021-07-07
Fix celery beat deployment connection to pg bouncer

fuziontech:
Celery beat should connect to PG via pgbouncer


macobo (approved):
Oopsie.

I wonder if there's a way tests can catch this. celery beat is a weird service.

PR #39, opened 2021-07-08
Fix redirectToTLS usage in helm templates and add to values

tiina303:
2 things: fix the way redirectToTLS is used (needs to not be quoted in the define) & adds it to values (so it's documented in our ALL_VALUES readme etc).

Fixes the way https://github.com/PostHog/charts-clickhouse/pull/29 works. Specifically currently we never add the annotations
```
> cat ../values.yaml && helm template posthog charts/posthog -f ../values.yaml | grep -A13 "# Source: posthog/templates/ingress.yaml"
cloud: "do"
ingress:
  type: "nginx"
# Source: posthog/templates/ingress.yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
 name: posthog
 labels:
    app: posthog
    chart: "posthog-1.4.36"
    release: "posthog"
    heritage: "Helm"
 annotations:
spec:
  tls:
  - hosts:

> cat ../values.yaml && helm template posthog charts/posthog -f ../values.yaml | grep -A13 "# Source: posthog/templates/ingress.yaml"
cloud: "do"
ingress:
  type: "nginx"
  redirectToTLS: true
# Source: posthog/templates/ingress.yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
 name: posthog
 labels:
    app: posthog
    chart: "posthog-1.4.36"
    release: "posthog"
    heritage: "Helm"
 annotations:
spec:
  tls:
  - hosts:

> cat ../values.yaml && helm template posthog charts/posthog -f ../values.yaml | grep -A13 "# Source: posthog/templates/ingress.yaml"
cloud: "do"
ingress:
  type: "nginx"
  redirectToTLS: false
# Source: posthog/templates/ingress.yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
 name: posthog
 labels:
    app: posthog
    chart: "posthog-1.4.36"
    release: "posthog"
    heritage: "Helm"
 annotations:
spec:
  tls:
  - hosts:
```

After this diff annos are added either if it's explicitly set to true or if not defined
```
> cat ../values.yaml && helm template posthog charts/posthog -f ../values.yaml | grep -A13 "# Source: posthog/templates/ingress.yaml"
cloud: "do"
ingress:
  type: "nginx"
# Source: posthog/templates/ingress.yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
 name: posthog
 labels:
    app: posthog
    chart: "posthog-1.4.36"
    release: "posthog"
    heritage: "Helm"
 annotations:
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
    cert-manager.io/cluster-issuer: "letsencrypt-prod"
 

> cat ../values.yaml && helm template posthog charts/posthog -f ../values.yaml | grep -A13 "# Source: posthog/templates/ingress.yaml"
cloud: "do"
ingress:
  type: "nginx"
  redirectToTLS: true
# Source: posthog/templates/ingress.yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
 name: posthog
 labels:
    app: posthog
    chart: "posthog-1.4.36"
    release: "posthog"
    heritage: "Helm"
 annotations:
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
    cert-manager.io/cluster-issuer: "letsencrypt-prod"


> cat ../values.yaml && helm template posthog charts/posthog -f ../values.yaml | grep -A13 "# Source: posthog/templates/ingress.yaml"
cloud: "do"
ingress:
  type: "nginx"
  redirectToTLS: false
# Source: posthog/templates/ingress.yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
 name: posthog
 labels:
    app: posthog
    chart: "posthog-1.4.36"
    release: "posthog"
    heritage: "Helm"
 annotations:
spec:
  tls:
  - hosts:
```

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -271,8 +271,8 @@ ingress:
   nginx:
     # -- Whether nginx is enabled
     enabled: true
-  # -- Whether to redirect to TLS. Defaults to true.
-  redirectToTLS:
+  # -- Whether to redirect to TLS with nginx ingress.
+  redirectToTLS: true
 
 postgresql:
   # -- Install postgres server on kubernetes (see below)
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -271,8 +271,8 @@ ingress:
   nginx:
     # -- Whether nginx is enabled
     enabled: true
-  # -- Whether to redirect to TLS with nginx ingress.
-  redirectToTLS: true
+    # -- Whether to redirect to TLS with nginx ingress.
+    redirectToTLS: true
 
 postgresql:
   # -- Install postgres server on kubernetes (see below)
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -271,6 +271,8 @@ ingress:
   nginx:
     # -- Whether nginx is enabled
     enabled: true
+  # -- Whether to redirect to TLS. Defaults to true.
+  redirectToTLS:
 
 postgresql:
   # -- Install postgres server on kubernetes (see below)
--- a/charts/posthog/templates/_helpers.tpl
+++ b/charts/posthog/templates/_helpers.tpl
@@ -344,11 +344,7 @@ Create the name of the service account to use
 Should we redirect HTTP to TLS?
 */}}
 {{- define "ingress.redirectToTLS" -}}
-{{- if ne (.Values.ingress.redirectToTLS | toString) "<nil>" -}}
     {{ .Values.ingress.redirectToTLS }}
-{{- else -}}
-    true
-{{- end -}}
 {{- end -}}
 
 
--- a/charts/posthog/templates/_helpers.tpl
+++ b/charts/posthog/templates/_helpers.tpl
@@ -344,10 +344,10 @@ Create the name of the service account to use
 Should we redirect HTTP to TLS?
 */}}
 {{- define "ingress.redirectToTLS" -}}
-{{- if hasKey .Values.ingress "redirectToTLS" -}}
-    {{ .Values.ingress.redirectToTLS | quote }}
+{{- if ne (.Values.ingress.redirectToTLS | toString) "<nil>" -}}
+    {{ .Values.ingress.redirectToTLS }}
 {{- else -}}
-    "true"
+    true
 {{- end -}}
 {{- end -}}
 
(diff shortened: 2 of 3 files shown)


charts/posthog/values.yaml
```diff
@@ -271,6 +271,8 @@ ingress:
   nginx:
     # -- Whether nginx is enabled
     enabled: true
+  # -- Whether to redirect to TLS. Defaults to true.
+  redirectToTLS:
```
macobo, line 275: Why do this in such a roundabout way (needing a custom helper)? 

Why not set it to true here and remove the code in _helpers?

macobo:
Is there a way to add tests catching this?

charts/posthog/values.yaml
tiina303: Good point, will change it.

tiina303:
> Is there a way to add tests catching this?

Likely starting to use https://github.com/helm/chart-testing or something similar would be the most reasonable way

charts/posthog/values.yaml
tiina303: left it in helpers template atm (instead of changing ingress.yaml to use .Values), we might want to have more complex logic later

macobo:
Not sure I understand. There's already a test suite which spins up a kubernetes cluster. Chart-testing provides no tooling on top of that except maybe for checking the output yaml.

I guess my question is - can we extend the existing suite in a way that would demonstrate the bug here? 

This is not a requirement for merge but for sake of keeping quality high going forward.

charts/posthog/values.yaml
macobo: We can always refactor that easily. YAGNI

macobo:
Pushed a patch:
- The behavior is nginx-specific so putting it under ingress.nginx
- Removing the helper - less layers of abstractions = win

tiina303:
Thanks for the update. Good to go from my perspective.

> can we extend the existing suite in a way that would demonstrate the bug here?

I can only think of a way to make it specific to the `ingress.type` parameter with the existing tests (e.g. we can add a `fail` to the template if `ingress.type` is null in the end). For something non-specific to this param: that's why I linked ct, which looked like it could display diffs and other, but I didn't dig into it too deep.

PR #40, opened 2021-07-08
Make tests less flaky

macobo:
Tests sometimes fail at exactly 10m of waiting for deployments to start
due to

```
Error: Action failure: rollout of deploy/posthog-web
The process '/usr/local/bin/kubectl' failed with exit code 1

error: deployment "posthog-web" exceeded its progress deadline
```

This is caused by a default in kubernetes - each deployment has
`spec.progressDeadlineSeconds` set to 600.

This PR increases that threshold to 1200 during tests.

This deadline seems to only affects `kubectl rollout` commands, so no
production effects here.

### DIFF
--- a/charts/posthog/templates/beat-deployment.yaml
+++ b/charts/posthog/templates/beat-deployment.yaml
@@ -8,7 +8,7 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 spec:
-  progressDeadlineSeconds: {{ .Values.worker.progressDeadlineSeconds }}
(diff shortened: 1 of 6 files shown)


macobo:
Marked as a draft - seems like issue is not timeouts but something not starting up cleanly sometimes. (Migrations?)

tiina303:
> Marked as a draft - seems like issue is not timeouts but something not starting up cleanly sometimes. (Migrations?)

Potentially could the problem be that we need the migrations to be done before something can start up nicely? in which case increasing the timeout would make sense.

PR #41, opened 2021-07-08
Update Readme adding placeholders for AWS and other install options

tiina303:
The readme currently was focused on GCP, tried to make it cover more ground + various nits. Will try to do the actual deployments on GCP and AWS too & document further (and add a bit more details for DO setup where I have manually deployed to already).

### DIFF
--- a/charts/posthog/README.md
+++ b/charts/posthog/README.md
@@ -6,8 +6,6 @@
 [![Slack](https://img.shields.io/badge/PostHog_chat-slack-blue?logo=slack)](https://posthog.com/slack)
 
 
-### :warning: This chart is still under development! Proceed with caution.
-
 -----
 
 🦔 [PostHog](https://posthog.com/) is developer-friendly, open-source product analytics.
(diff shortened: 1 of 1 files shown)


macobo:
What should I as a reviewer review here?

charts/posthog/values.yaml
```diff
@@ -14,7 +14,7 @@ image:
   pullPolicy: IfNotPresent
 
 # -- Cloud service being deployed on. Either `gcp` or `aws` or `do` for DigitalOcean
-cloud: "gcp"
+cloud:
```
macobo, line 17: This is not a README change.

charts/posthog/templates/ingress.yaml
```diff
@@ -17,16 +17,6 @@ metadata:
     kubernetes.io/ingress.global-static-ip-name: {{ .Values.ingress.gcp.ip_name | quote }}
     {{- end }}
    {{- end }}
-   {{- if (eq (include "ingress.type" .) "alb") }}
```
macobo, line 20: Why remove this as part of README changes?

charts/posthog/README.md
```diff
@@ -283,9 +379,29 @@ If running the command shows the SSL cert as `PROVISIONING`, that means that the
...
+web:
+  secureCookies: false
+```
+
+</details>
+
 ### How can I connect to my ClickHouse instance?
 
-- Get the IP via `kubectl get svc`
+- Get the IP via `kubectl get svc --namespace posthog`
```
macobo, line 404: This change is wrong: Other commands here don't encourage specifying the namespace. Users can also specify any namespace they'd like - `--namespace posthog` varies from one to the other.

charts/posthog/README.md
```diff
@@ -127,7 +205,7 @@ postgresql:
 
 redis:
   master:
-    size: 2Gi
+    size: 10Gi
```
macobo, line 208: Why?

charts/posthog/README.md
```diff
@@ -143,31 +221,21 @@ worker:
 
 plugins:
   replicacount: 2
-
-pgbouncer:
```
macobo, line 147: Why?

charts/posthog/README.md
```diff
@@ -100,24 +182,20 @@ Dependent charts can also have values overwritten. Preface values with postgresq
...
 
-This `values.yaml` is geared towards minimizing cost on google kubernetes engine. Use it if you have less than 1M events/month.
-
 ```yaml
-# Minimal settings for scale of 10k to 1M events/month.
 # Note that this is experimental, please let us know how this worked for you.
 
-cloud: gcp
-ingress:
-  hostname: "posthog.yourcloud.net"
```
macobo, line 119: Why? Without this users won't be able to access their cluster.

charts/posthog/README.md
```diff
@@ -100,24 +182,20 @@ Dependent charts can also have values overwritten. Preface values with postgresq
...
 
 <br />
 
-This `values.yaml` is geared towards minimizing cost on google kubernetes engine. Use it if you have less than 1M events/month.
-
 ```yaml
-# Minimal settings for scale of 10k to 1M events/month.
 # Note that this is experimental, please let us know how this worked for you.
 
-cloud: gcp
```
macobo, line 117: Why? Without this this yaml file does not work.

charts/posthog/README.md
```diff
@@ -16,47 +16,129 @@ This chart bootstraps a [PostHog](https://posthog.com/) deployment on a [Kuberne
...
+  user: <your STMP login user>
+  password:  <your STMP password>
+```
+
 ## Upgrading the chart
 
 ```console
 helm repo update
-helm upgrade -f values.yaml posthog posthog/posthog
+helm upgrade -f values.yaml --timeout 20m --namespace posthog posthog posthog/posthog
```
macobo, line 141: Why set namespace here? This only complicates instructions.

Re upgrade timeout: Why is it needed? Timeout is needed during initial install since clickhouse-operator takes a while to setup. This is not the case with upgrades.

charts/posthog/README.md
```diff
@@ -16,47 +16,129 @@ This chart bootstraps a [PostHog](https://posthog.com/) deployment on a [Kuberne
...
 After the chart has started, you can look up the external ip via the following command:
 
 ```bash
-kubectl get svc posthog
+kubectl get svc posthog --namespace posthog
 ```
 
 Create a `CNAME` record from your desired hostname to the external IP.
 
+### Setting up email
```
macobo, line 129: Structure-wise this is awkwardly placed. This should be under the `configuration` section as everything else.

charts/posthog/README.md
```diff
@@ -16,47 +16,129 @@ This chart bootstraps a [PostHog](https://posthog.com/) deployment on a [Kuberne
...
 ## Installing the chart
 
-To install the chart with the release name `posthog`, run the following:
+To install the chart with the release name `posthog` in `posthog` namespace, run the following:
 
 ```console
 helm repo add posthog https://posthog.github.io/charts-clickhouse/
 helm repo update
-helm install -f values.yaml --timeout 20m posthog posthog/posthog
+helm install -f values.yaml --timeout 20m --create-namespace --namespace posthog posthog posthog/posthog
```
macobo, line 55: Why install in a namespace by default in the instructions here.

I think we can assume the users who want separate namespaces will know how to do so. Encouraging to do so by default:
- increases command complexity, the amount of flags you need to specify when installing
- It makes debugging _anything_ on the cluster more complicated.

If you're installing on a shared cluster, sure that makes sense. However I don't think we should encourage it here.

charts/posthog/README.md
```diff
@@ -16,47 +16,129 @@ This chart bootstraps a [PostHog](https://posthog.com/) deployment on a [Kuberne
...
+```
+
+Where the `values.yaml` has the following minimum content:
```
macobo, line 58: This structure is kind of bad. We now have install section which gives some overview of configuration and then a separate configuration section.

I'd recommend keeping all of configuration in one place and _linking_ to relevant sections instead. As-is this just makes the flow harder to follow.

charts/posthog/templates/ingress.yaml
tiina303: being lazy ...  because I was looking into this as I was writing up the minimum values files, but I can make it a separate PR.

charts/posthog/README.md
tiina303: I think if we always use a namespace it works the same way for everyone being on a shared cluster or not. It's also much easier for us to say just nuke the whole namespace with everything in it without needing to verify they are only running Posthog there first. There's an easy command for kubectl to be in a specific namespace, so you don't need to provide this argument every time. That said I don't feel strongly about this, just figured it could make our lives easier in the long run.

charts/posthog/README.md
tiina303: The way I saw it is "Configuration" is optional stuff, this in here is mandatory (so the user needing to go down to the configuration section added complexity for most folks, who just wanted to make the simplest thing run which I believe is the majority). I'd love if we didn't need it, but I haven't seen a way we could make the chart work with default values for multiple deployments. I looked into it and then only needing 1 value (cloud being the only required setting through values.yaml or `--set cloud=gcp` or `--set cloud=aws`), but being able to enable/disable nginx,certManager was tricky based on other values in the yaml sadly.

charts/posthog/README.md
tiina303: Sure this is kind of optional, so happy to move.

charts/posthog/README.md
tiina303: Can't slow migrations take longer than 5min sometimes (what about clickhouse version upgrades)?

charts/posthog/README.md
tiina303: This is gcp specific, I moved that part up to the minimal required configuration within install section, the sizing stuff is otherwise is platform agnostic.

charts/posthog/README.md
tiina303: same, this is part of the required config that I moved up to install

charts/posthog/README.md
tiina303: That's the default, so skip it makes things shorter/simpler

charts/posthog/README.md
tiina303: The default in our values is 5Gi, so this going smaller didn't make sense (given that this section is about scaling up further from default values). If we don't need to increase it to scale up I'd just remove this redis size change instead.

tiina303:
Still wip, but going to separate out per platform like Karl suggested.
We could consider changing the guide to start without having a hostname specified as folks might run into issues and things are faster if they don't need to figure out how to access it & wait for DNS to update etc. Then step 2 is securing it.

charts/posthog/README.md
macobo: I don't think we really need to scale this up at all. Redis is treated more as a ephemeral service by everything but plugins.

PR #45, opened 2021-07-12
Add default cluster name as posthog to clickhouse deployments

fuziontech:
This is a base requirement for changing how migrations work for clickhouse so we can use `ON CLUSTER` command for DDLs.

### DIFF
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -47,7 +47,7 @@ spec:
     profiles:
       default/allow_experimental_window_functions: "1"
     clusters:
-      - name: {{ .Values.clickhouse.database | quote }}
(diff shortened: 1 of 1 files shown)


tiina303:
What happens when someone upgrades their Posthog deployment and goes from no cluster name to posthog?

fuziontech:
Nothing should happen since it's just one node clusters and we are not using ZK yet or any installs. Better to do this now than later when we have larger installs.

I'm about to test applying this change to a test cluster on DO though to make sure.

tiina303:
& why do we care about using `ON CLUSTER` command if we just have one cluster everywhere? Do you think we'll be using ZK or multiple clusters soon?

fuziontech:
Cloud has > 1 node

fuziontech:
Actually we don't need this! 🎉

PR #49, opened 2021-07-13
Remove gcp as default cloud value

tiina303:
Currently our default settings have:
1. cloud=gcp
2. nginx and certManager enabled

Those two settings together don't make sense as on gcp we'd want to use their ingress controller, so we shouldn't enable nginx, on the other hand having cloud be gcp changes ingress configuration to not what we want when using nginx.

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -14,7 +14,7 @@ image:
   pullPolicy: IfNotPresent
 
 # -- Cloud service being deployed on. Either `gcp` or `aws` or `do` for DigitalOcean
-cloud: "gcp"
+cloud:
 # -- Sentry endpoint to send errors to
 sentryDSN:
 


macobo:
Wouldn't it make more sense to fix `nginx and certManager enabled by default`? This seems to have arisen from recent refactoring/extracting `"ingress.type"` helper?

That said, this looks fine.

fuziontech:
I agree could be cleaned up more, but considering that defaults don't work currently this seems like a reasonable first step

fuziontech (approved):

tiina303:
> Wouldn't it make more sense to fix `nginx and certManager enabled by default`? This seems to have arisen from recent refactoring/extracting "ingress.type" helper?

Regardless of this I think the cloud parameter should be empty by default. Put up a pr for changing nginx and certManager to default to false https://github.com/PostHog/charts-clickhouse/pull/51

PR #51, opened 2021-07-14
Disable nginx and certManager by default

tiina303:
## Changes

Disables nginx and certManager by default
Enable `letsencrypt` by default condition to additionally check that hostname exists and certManager is enabled

## Why?

It's odd to need to disable them for gcp + we might in the future want to use the [AWS Load Balancer Controller](https://github.com/kubernetes-sigs/aws-load-balancer-controller) (previously know as alb ingress controller)

## Ideal world

I wish we could enable/disable nginx and certManager based on the cloud variable str value (if gcp disable, if do enable both), but sadly it's not possible.
Specifically we can have boolean conditions to enable/disable or tags but not use template variables there nor expressions. Docs https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies

Based on re-reading the docs we could do sth like (note that the first path that exists is used):
```
name: nginx
  condition: ingress.nginx.enabled

name: certManager
  condition: certManager.enabled, ingress.nginx.enabled
```

Then in the values files we can by default not define certManager, but it's possible to disable it by `certManager.enabled: false`. That said this seems fragile as it's easy for someone to forgot that we're relying on the first path that exists is used and define `certManager.enabled` as empty string.

### DIFF
--- a/charts/posthog/templates/_helpers.tpl
+++ b/charts/posthog/templates/_helpers.tpl
@@ -360,7 +360,7 @@ Create the name of the service account to use
 {{- define "ingress.letsencrypt" -}}
 {{- if ne (.Values.ingress.letsencrypt | toString) "<nil>" -}}
   {{ .Values.ingress.letsencrypt }}
-{{- else if .Values.ingress.nginx.enabled -}}
+{{- else if and (and (.Values.ingress.nginx.enabled) (.Values.certManager.enabled)) (ne (.Values.ingress.hostname | toString) "<nil>")  -}}
   true
 {{- else -}}
   false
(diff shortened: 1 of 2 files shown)


fuziontech:
This makes sense, but what if we wanted to do something small like have a way to push the hostname to k8s for DO? I feel like we should leave it in there just to make it easier to enable.

tiina303:
Well the follow-up would be to change the DigitalOcean values file (where we can set cert-manager to true without using it immediately), so there's no difference for DigitalOcean deploys. Or did you mean that for `letsencrypt` enabling part?

fuziontech (approved):

PR #52, opened 2021-07-15
Add deploymentKind to helm info

tiina303:
Follow-up to this will be setting it to "DigitalOcean 1-Click" in the values file in their marketplace.

Verified that we get the info by checking recent `helm install` events [here](https://app.posthog.com/events)

Screen Shot 2021-07-15 at 02 28 18

### DIFF
--- a/charts/posthog/templates/_helpers.tpl
+++ b/charts/posthog/templates/_helpers.tpl
@@ -350,6 +350,9 @@ Create the name of the service account to use
 {{- $info := set $info "hostname" .Values.ingress.hostname -}}
 {{- $info := set $info "operation" (include "posthog.helmOperation" .) -}}
 {{- $info := set $info "ingress_type" (include "ingress.type" .) -}}
+{{- if .Values.deploymentKind -}}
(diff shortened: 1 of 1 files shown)


charts/posthog/templates/_helpers.tpl
```diff
@@ -350,6 +350,9 @@ Create the name of the service account to use
 {{- $info := set $info "hostname" .Values.ingress.hostname -}}
 {{- $info := set $info "operation" (include "posthog.helmOperation" .) -}}
 {{- $info := set $info "ingress_type" (include "ingress.type" .) -}}
+{{- if .Values.deploymentKind -}}
+{{- $info := set $info "deployment_kind" .Values.deploymentKind -}}
```
macobo, line 354: kind is kind of a bad name - overloaded

I suggest `medium` with default to `helm`

tiina303:
Updated https://github.com/digitalocean/marketplace-kubernetes/pull/239 accordingly.

What's the easiest way and should we remove the `deployment_kind` property?

PR #56, opened 2021-07-19
Fix deployment type defaulting in chart template

tiina303:
We were always sending "helm" as the deployment type because of the args ordering ... the default function `default DEFAULT_VALUE GIVEN_VALUE` [docs](https://helm.sh/docs/chart_template_guide/function_list/#default), but it's easier to just use [pipelines](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) everywhere.

Also changed the cloud value to "" so without providing it the chart compiles

before
```
> helm template posthog ../charts-clickhouse2/charts/posthog | tail
Error: template: posthog/templates/clickhouse_instance.yaml:2:7: executing "posthog/templates/clickhouse_instance.yaml" at <eq .Values.cloud "gcp">: error calling eq: incompatible types for comparison

Use --debug flag to render out invalid YAML

> helm template posthog ../charts-clickhouse2/charts/posthog --values ../marketplace-kubernetes/stacks/posthog/values.yml | tail
          value: "false"
        - name: DEFAULT_FROM_EMAIL
          value: "hey@posthog.com"
        - name: HELM_INSTALL_INFO
          value: "{\"chart_version\":\"2.0.1\",\"cloud\":\"do\",\"deployment_type\":\"helm\",\"hostname\":null,\"ingress_type\":\"nginx\",\"operation\":\"install\",\"release_name\":\"posthog\",\"release_revision\":1}"
        resources:
          limits:
            memory: 1000Mi
          requests:
            memory: 1000Mi
```

After
```
helm template posthog charts/posthog | tail
          value: "false"
        - name: DEFAULT_FROM_EMAIL
          value: "hey@posthog.com"
        - name: HELM_INSTALL_INFO
          value: "{\"chart_version\":\"2.0.1\",\"cloud\":\"\",\"deployment_type\":\"helm\",\"hostname\":null,\"ingress_type\":\"nginx\",\"operation\":\"install\",\"release_name\":\"posthog\",\"release_revision\":1}"
        resources:
          limits:
            memory: 1000Mi
          requests:
            memory: 1000Mi

> helm template posthog charts/posthog --values ../marketplace-kubernetes/stacks/posthog/values.yml | tail
          value: "false"
        - name: DEFAULT_FROM_EMAIL
          value: "hey@posthog.com"
        - name: HELM_INSTALL_INFO
          value: "{\"chart_version\":\"2.0.1\",\"cloud\":\"do\",\"deployment_type\":\"DigitalOcean 1-click\",\"hostname\":null,\"ingress_type\":\"nginx\",\"operation\":\"install\",\"release_name\":\"posthog\",\"release_revision\":1}"
        resources:
          limits:
            memory: 1000Mi
          requests:
            memory: 1000Mi
```

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -14,7 +14,7 @@ image:
   pullPolicy: IfNotPresent
 
 # -- Cloud service being deployed on. Either `gcp` or `aws` or `do` for DigitalOcean
-cloud:
+cloud: ""
 # -- Sentry endpoint to send errors to
 sentryDSN:
 
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -14,7 +14,7 @@ image:
   pullPolicy: IfNotPresent
 
 # -- Cloud service being deployed on. Either `gcp` or `aws` or `do` for DigitalOcean
-cloud: ""
+cloud:
 # -- Sentry endpoint to send errors to
 sentryDSN:
 
--- a/charts/posthog/templates/_helpers.tpl
+++ b/charts/posthog/templates/_helpers.tpl
@@ -350,7 +350,7 @@ Create the name of the service account to use
 {{- $info := set $info "hostname" .Values.ingress.hostname -}}
 {{- $info := set $info "operation" (include "posthog.helmOperation" .) -}}
 {{- $info := set $info "ingress_type" (include "ingress.type" .) -}}
-{{- $info := set $info "deployment_type" (default .Values.deploymentType "helm") -}}
+{{- $info := set $info "deployment_type" (.Values.deploymentType | default "helm") -}}
 {{ toJson $info | quote }}
 {{- end -}}
 


macobo:
> Also changed the cloud value to "" so without providing it the chart compiles

Failing loudly is failing better than failing late in the process. You removed the default in a previous PR and without setting it the chart _will not work_ - we should either add back gcp or something else as a default OR fail loudly as we do currently.

fuziontech:
I generally agree with @macobo - We should have cloud be a required parameter since it is so important to how we configure things down the line with block storage, services, and ingress

tiina303:
> You removed the default in a previous PR and without setting it the chart will not work 

"without setting it the chart will not work " this is only the case because we have a bunch of `eq` checks against a string. where we use `eq .Values.cloud <string>` that errors if `cloud` isn't set, what we should be doing there is checking that cloud is set and then compare. These eq checks work fine if `cloud = ""`. Currently when deploying on DigitalOcean we set the cloud value to "do" but don't use it anywhere, similarly if someone wants to deploy on their own data centers we wouldn't want to have the cloud value change anything, so an empty or "" or "\<name-not-used-anywhere-else\>" would be a fine default for those use cases.

So options:
1. complain loudly if cloud isn't set (we can do this via the required parameter https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function)
2. add back gcp - this seems like a bad idea to me as it does things that are gcp specific and won't work anywhere else
3. add something else as default - I like the idea of default setup being as cloud agnostic as possible and working on bare metal as well as possible (instead of empty str it could be "bare-metal" or "undefined"). I'd also like to be able to run `helm template` without needing to provide it any values.
4. fix the eq comparisons and we can leave it like it is now. I assume it's possible, but didn't check.

I like option 3 the most, not sure what it should be named.

macobo:
I'd start from the users perspective here. 

1. It's better to fail loudly than with a whimper N steps into a deploy
2. We should minimize code around plumbing if possible

Given we only support 3 options for cloud right now adding a 4th (e.g. empty or bare-metal, etc) that doesn't really work is a bad idea - we're just increasing the support burden everywhere.

>  I'd also like to be able to run helm template without needing to provide it any values.

Why? If the chart would not be deployable without values then helm template without values is useless. Just pass in the values with `-f` you're testing against.

Given this increases the complexity of the chart and works against other goals (failing loudly, maintainability) this is _not_ something to work towards.

--- 

All-in-all what is the minimum change that will work for the users? It's easy for you to change the command you're running, but code changes and user breakages are much more expensive.

tiina303:
> Given we only support 3 options for cloud right now adding a 4th (e.g. empty or bare-metal, etc) that doesn't really work is a bad idea - we're just increasing the support burden everywhere.

What do you mean "doesn't really work". The "do" cloud option does nothing. So it's the same as if we had empty str or bare-metal & it works fine. Same way someone deploying to bare metal would work fine (well depends on their internals of course, but if it can work fine on DigitalOcean it can work fine on bare metal too).

> Why? If the chart would not be deployable without values then helm template without values is useless

It is deployable without values, see above.

From a user perspective: 3: why do I need to provide something that isn't used.

The reason to require cloud parameter is your (1) point, but to elaborate because the user is likely using AWS or GCP and most likely forgot the cloud parameter, since there are less users off those platforms it's a fine tradeoff to yell at them all to say cloud=something

charts/posthog/values.yaml
```diff
@@ -14,7 +14,7 @@ image:
   pullPolicy: IfNotPresent
 
 # -- Cloud service being deployed on. Either `gcp` or `aws` or `do` for DigitalOcean
-cloud:
+cloud: ""
```
tiina303, line 17:
```suggestion
cloud:
```

PR #58, opened 2021-07-20
Made resource requirements for DO 1 click clearer

marcushyett-ph:
Make resource requirements for DO 1 click clearer

### DIFF
--- a/charts/posthog/README.md
+++ b/charts/posthog/README.md
@@ -142,6 +142,8 @@ web:
     <b>Digital Ocean</b>
   </summary>
 <br />
+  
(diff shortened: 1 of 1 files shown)


tiina303:
Hmm...I kinda want the link to 1-click be first & it should all be self documenting there (in 2 places: DO marketplace app install page & posthog.com deployment page), which it currently isn't clearly. 
Why: I see the audience of the readme to be 
1) folks who got lost & should be redirected to 1-click docs/deployment 
2) folks wanting to use helm directly

That said happy to land it now while I haven't gotten all the docs in order.

PR #61, opened 2021-07-20
Add troubleshooting section about helm install failure due to not enough resources

tiina303:
Also removed the prerequisites that confused Marcus in https://docs.google.com/document/d/1bzgs5TQri85gYZSHWRQNON49MfdVMQ19SRw77X23qNE/edit# to be inlined, especially since each section describes how to get the Kubernetes cluster


macobo:
Information architecture feedback: We previously had a trouble-shooting section, but afaik you moved all of then under individual `installing the chart` sections. Why have this separately then?

fuziontech (approved):
lgtm :shipit:

tiina303:
> Information architecture feedback: We previously had a trouble-shooting section, but afaik you moved all of then under individual installing the chart sections. Why have this separately then?

Can't make up my mind where things should be yet :D

PR #63, opened 2021-07-21
Fix migrations by adding zookeeper for clickhouse

tiina303:
Migrations failed with a message
```
infi.clickhouse_orm.database.ServerError: There is no Zookeeper configuration in server config (version 21.6.5.37 (official build)) (139)
```

which happened because we were running 
```
DROP TABLE ... ON CLUSTER posthog
```
the `ON CLUSTER` part requires zookeeper.

Longer term we should move the Kafka zookeeper to combine with this one as we don't need to run 2 ZKs.

TODO:
- [x] fix values
- [x] test on DO
  - [x] upgrade 
  - [x] install (couldn't create a new cluster, so tried in a new namespace 
```
helm install -f ~/Posthog/tiina-values/release-testing-do.yaml posthog ~/Posthog/charts-clickhouse2/charts/posthog --namespace posthog2 --create-namespace --set image.tag=latest --set clickhouseOperator.namespace=posthog2
```
- [x] template around clickhouse operator enabled

## NOTE & WARNING
Because this [alters the kafka zk cluster](https://stackoverflow.com/questions/59592518/kafka-broker-doesnt-find-cluster-id-and-creates-new-one-after-docker-restart) you will have to shell into the kafka node and remove `bitnami/kafka/data/meta.properties` file

```console
rm bitnami/kafka/data/meta.properties
```

This can be difficult so what is easier is to delete two things and re-deploy
- delete kafka stateful set
- delete kafka pods
- upgrade helm chart


macobo:
Q: Why do we need 2 of zk? It's gonna be hard to migrate to 1 afterwards so why not wire up ch to 1 right now?

tiina303:
> Q: Why do we need 2 of zk? It's gonna be hard to migrate to 1 afterwards so why not wire up ch to 1 right now?

We don't. But the one we have now is a dependent one from Kafka, we don't want that to be the one we will have in the end (Kafka could get rid of the dep). So the idea is to create one generic one for posthog and use it for both Clickhouse and Kafka. But given this is blocking updates the fastest thing was to create it first separately & then move kafka zk usage over later.

macobo:
Let's think about this from the users perspective.

This chart now spends extra resources spinning up independent duplicates of a mission-critical service. Both of whom are managed differently, versioned differently and have different needs. This sounds like hell.

Give n zk needs to be an available and durable data store migrating later is bound to cause downtime and issues for them as well - making the experience using the chart worse.

On the other end we have potential maintenance overhead of kafka chart, whose versioning we fully control and where zk removal is not gonna be relevant for a year if not more.

YAGNI

tiina303:
Note2: we just took the values from https://github.com/Altinity/clickhouse-operator/blob/master/deploy/zookeeper/quick-start-persistent-volume/zookeeper-1-node.yaml and if we were to use the Kafka ZK we might have configuration there that doesn't work all the great for Clickhouse.

I don't have strong opinions here, as I don't have enough insights/experience in this area: e.g. how easy it would be to move the Kafka zk usage under this generic one, what problems could happen if we use the kafka zk, ...

charts/posthog/templates/_helpers.tpl
```diff
@@ -132,6 +132,20 @@ Set postgres URL
     postgres://{{- .Values.postgresql.postgresqlUsername -}}:{{- template "posthog.postgresql.password" . -}}@{{- template "posthog.postgresql.host" .  -}}:{{- template "posthog.postgresql.port" . -}}/{{- .Values.postgresql.postgresqlDatabase }}
 {{- end -}}
 
+{{/*
+Set zookeeper host
+*/}}
+{{- define "posthog.zookeeper.host" -}}
+    {{- template "posthog.fullname" . }}-posthog-zookeeper
```
tiina303, line 139: doesn't `posthog.fullname` already contain `posthog`?

fuziontech: It does, but if you look at how we name all of the dependent services they end up `posthog-posthog-<service>` which is the `<deployment>-posthog-<service>` which works out, though a bit repetitive usually.

fuziontech (approved):

jredl-va:
Thanks Team!

twobitfool:
@tiina303 I'm running into the exact same error...

```
infi.clickhouse_orm.database.ServerError: There is no Zookeeper configuration in server config (version 21.6.5.37 (official build)) (139)
```

...when trying to upgrade from chart posthog-3.15.1 to posthog-3.16.0, but I'm not clear on how to resolve it. Any help would be much appreciated.

tiina303:
Why are you trying to update to posthog-3.16.0 instead of the current latest chart version?

twobitfool:
@tiina303: I want to upgrade to the latest version, but when I tried, I ran into several errors. So, I was going with **Plan B**: Install a fresh copy of PostHog (latest version), and then [migrate the old data](https://posthog.com/docs/self-host/migrate/migrate-to-another-self-hosted-instance).

...but the data export process [requires PostHog v1.30.0](https://posthog.com/docs/self-host/migrate/migrate-to-another-self-hosted-instance#requirements)

tiina303:
For plan A: did you see the same zk error?
For plan B: try not upgrading the chart, but just the posthog version (`image.tag = "release-1.30.0"`)

tiina303:
Looks like the major upgrade notes likely were not done for https://posthog.com/docs/self-host/deploy/upgrade-notes#upgrading-from-2xx

PR #68, opened 2021-07-26
Add faq about increasing storage

tiina303:
I tested that it worked on DigitalOcean (checked data was there & storage changed via `SELECT * FROM system.disks;` on Clickhouse) & James mentioned that it works on AWS & GCP


fuziontech (approved):
I'm sure we would have been asked about this. Nice.

Issue #47, opened 2021-07-13
Digital Ocean deployment post-install


tiina303:
## Is your feature request related to a problem?

Currently the Digital Ocean 1-click install button deploys Posthog with TLS disabled. Furthermore the update button would likely revert any changes made.
Note to self: Wordpress has a section about configuring hostname and LetsEncrypt: https://marketplace.digitalocean.com/apps/wordpress (though oddly no info about that in their kubernetes one https://marketplace.digitalocean.com/apps/wordpress-kubernetes)

Also migrations might not be able to allocate (not enough memory)

## Describe the solution you'd like

Description to users how to secure Posthog and update to configs so that the 1-click upgrades don't revert the changes.

## Describe alternatives you've considered

Using parameters, but sadly DigitalOcean doesn't support it at this point.

tiina303:
Since the update button doesn't currently do anything I'm deprioritizing restructuring our values to use in the hopes that they add parameters beforehand and we don't need to do it.

tiina303:
Updated the getting started part in the DigitalOcean marketplace too, which is now live also here: https://marketplace.digitalocean.com/apps/posthog-1

PR #70, opened 2021-07-26
[readme] Add 1-click DigitalOcean post-install steps for enabling TLS

tiina303:
https://github.com/PostHog/charts-clickhouse/issues/47

Will add the same instructions to the [marketplace page](https://marketplace.digitalocean.com/apps/posthog-1) afterwards .

Also fixed the redirectToTLS value location


fuziontech (approved):
nice, this is great! Thanks for making sure the TLS redirect is handled after setting up TLS

PR #72, opened 2021-07-27
Update chart to use new app version by default

tiina303:
Did a quick check that install on DigitalOcean worked fine.


fuziontech (approved):
shipit

PR #73, opened 2021-07-27
Update instructions for upgrading to v3

tiina303:
The pods get deleted if stateful sets get deleted and they get recreated otherwise, so the best strategy here seems to be 
1) delete kafka & zookeeper sts (so the pods won't restart)
2) delete kafka pvc (without this we still ran into the error about cluster ID) 
3) make sure kafka & zk pods shut down
4) run the upgrade.


fuziontech (approved):
lgtm

PR #74, opened 2021-07-27
add cloud param to other cloud

tiina303:
it errors out otherwise with cloud str comparison


fuziontech (approved):
seems worth of a bump in patch version :P

Issue #75, opened 2021-07-28
Install fails on non-posthog namespace


taobojlen:
## Bug description

I saw that your Clickhouse charts were now public, so I thought I'd have a go at installing them in our staging namespace to see how it works. I'm running into an issue, though, where the chart seems to expect the namespace to be `posthog`.

## Expected behavior

I can install the chart in any namespace.

## How to reproduce

1. `helm upgrade -i --timeout 20m --namespace main-tools-staging -f values-staging.yaml staging-posthog-clickhouse path/to/my/chart`
2. Get the error `Error: UPGRADE FAILED: failed to create resource: namespaces "posthog" not found`

## My chart
```
apiVersion: v2
name: posthog-clickhouse
type: application
version: 0.1.0
appVersion: "1.26.0"
dependencies:
  - name: posthog
    version: 3.0.0
    repository: https://posthog.github.io/charts-clickhouse
```

## Values
```
posthog-clickhouse:
  cloud: "gcp"
  certManager:
    enabled: false
  ingress:
    nginx:
      enabled: false
    hostname: posthog-clickhouse.staging.example.com
  hooks:
    migrate:
      hookAnnotation: "pre-install,pre-upgrade"
  clickhouseOperator:
    namespace: "main-tools-staging"
```

## Environment

- [x] Deployment platform (gcp/aws/...): GCP
- [x] Chart version/commit: 3.0.0
- [x] Posthog version: 1.26

macobo:
To fix, change `clickhouseOperator.namespace: "main-tools-staging"` in values.yaml

cc @fuziontech @tiina303 - having this option still seems slightly weird. Perhaps it should just default to the same namespace?

taobojlen:
The error still appears without setting `clickhouseOperator.namespace` (I actually only changed that in an attempt to work around the bug). Sorry, could've been clearer about that!

tiina303:
> The error still appears without setting clickhouseOperator.namespace (I actually only changed that in an attempt to work around the bug). Sorry, could've been clearer about that!

Sorry this is inconvenient. Currently you need to set that value or have "posthog" namespace. I agree that this is confusing and annoying that we need to define it, I'll look into removing that.

Why you're running into the error with your initial values that you provided: I'm not sure how your values are used as normally they would be at the top level, i.e.
instead of 
```yaml
posthog-clickhouse:
  cloud: "gcp"
  ...
```
you'd have
```yaml
cloud: "gcp"
...
```

[...]

PR #76, opened 2021-07-28
Fix comparisons of incompatible types

taobojlen:
I had to make these changes to the chart to avoid errors like the following:
```
Error: UPGRADE FAILED: template: posthog-clickhouse/charts/posthog/templates/clickhouse_instance.yaml:2:7: executing "posthog-clickhouse/charts/posthog/templates/clickhouse_instance.yaml" at <eq .Values.cloud "gcp">: error calling eq: incompatible types for comparison
```

This is using the following tools:
- helm 3.5.2
- kubectl 1.17.17
- kubernetes v1.19.10-gke.1600


taobojlen:
Closing -- the issue was resolved in https://github.com/PostHog/charts-clickhouse/issues/75#issuecomment-888443603

PR #77, opened 2021-07-28
Add FAQ about namespace posthog not found

tiina303:
https://github.com/PostHog/charts-clickhouse/issues/75


tiina303:
https://github.com/PostHog/charts-clickhouse/pull/79 is much better

PR #78, opened 2021-07-28
Make the error nicer when cloud isn't provided

tiina303:
Just making the error nicer if cloud parameter isn't defined. Note the required function needs to be executed for it to actually error out and this seemed like the best way to accomplish it.

before
```
> helm template posthog ~/Posthog/charts-clickhouse2/charts/posthog
Error: template: posthog/templates/clickhouse_instance.yaml:2:7: executing "posthog/templates/clickhouse_instance.yaml" at <eq .Values.cloud "gcp">: error calling eq: incompatible types for comparison
```
after
```
> helm template posthog ~/Posthog/charts-clickhouse/charts/posthog
Error: execution error at (posthog/templates/_helpers.tpl:362:32): 'cloud' value is required, e.g. 'gcp', 'aws', 'do', 'private', ...
```

We could also not require the `cloud` parameter, but the likelihood is that someone didn't provide it when they should have like happened in https://github.com/PostHog/charts-clickhouse/pull/76
There's also more discussion on how yelling loudly and failing early is better than deploying something that doesn't work tied to the cloud value specifically in  https://github.com/PostHog/charts-clickhouse/pull/56

### DIFF
--- a/charts/posthog/templates/_helpers.tpl
+++ b/charts/posthog/templates/_helpers.tpl
@@ -353,13 +353,10 @@ Create the name of the service account to use
 {{- end -}}
 {{- end -}}
 
-{{- define "cloud" -}}
-{{- required "'cloud' value is required, e.g. 'gcp', 'aws', 'do', 'private', ..." .Values.cloud -}}
-{{- end -}}
 
 {{- define "posthog.helmInstallInfo" -}}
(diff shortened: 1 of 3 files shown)


fuziontech (approved):
Seems like a good idea to require this

PR #79, opened 2021-07-28
Default CHOP namespace to the release namespace

fuziontech:
fixes https://github.com/PostHog/charts-clickhouse/issues/75﻿

### DIFF
--- a/charts/posthog/templates/clickhouse_cluster_role.yaml
+++ b/charts/posthog/templates/clickhouse_cluster_role.yaml
@@ -124,5 +124,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: clickhouse-operator
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }}
+  namespace: {{ .Release.Namespace }}
 {{- end }}
--- a/charts/posthog/templates/clickhouse_cluster_role.yaml
+++ b/charts/posthog/templates/clickhouse_cluster_role.yaml
@@ -124,5 +124,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: clickhouse-operator
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }} 
 {{- end }}
--- a/charts/posthog/templates/clickhouse_cluster_role.yaml
+++ b/charts/posthog/templates/clickhouse_cluster_role.yaml
@@ -124,5 +124,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: clickhouse-operator
-  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }} 
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }} 
 {{- end }}
--- a/charts/posthog/templates/clickhouse_config_map.yaml
+++ b/charts/posthog/templates/clickhouse_config_map.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etc-clickhouse-operator-files
-  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }} 
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }} 
   labels:
     app: clickhouse-operator
 data:
@@ -134,7 +134,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etc-clickhouse-operator-confd-files
-  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
   labels:
     app: clickhouse-operator
 data:
@@ -143,7 +143,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etc-clickhouse-operator-configd-files
-  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
   labels:
     app: clickhouse-operator
 data:
@@ -193,7 +193,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etc-clickhouse-operator-templatesd-files
-  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
   labels:
     app: clickhouse-operator
 data:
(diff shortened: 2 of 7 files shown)


macobo:
Wouldn't this break some existing installations since clickhouseOperator would move?

fuziontech:
> Wouldn't this break some existing installations since clickhouseOperator would move?

Only if they were installing this off label from our instructions would this be a problem. We have them install this in the `posthog` namespace in all of our instructions. Doing this now vs later would only make this transition cleaner and less painful.

> A more sane way is changing the default value to be blank and using `default .Values.clickhouseOperator.namespace .Release.namespace`

I would say _safe_ over _sane_ to be more accurate. The correct syntax would be `{{ .Values.clickhouseOperator.namespace | default .Release.namespace }}`

> In both cases, going strictly by semver this is a breaking change.

This would be breaking, but the remedy would be simply to have existing installations add `.Values.clickhouseOperator.namespace` back to their provided values.yaml file.

charts/posthog/values.yaml
```diff
@@ -21,8 +21,6 @@ sentryDSN:
 clickhouseOperator:
   # -- Whether to install clickhouse. If false, `clickhouse.host` must be set
   enabled: true
-  # -- Which namespace to install clickhouse operator to
-  namespace: "posthog"
```
tiina303, line 25: keep it just as empty, so we have the description. I believe the default function still works then

tiina303:
lets yes do the defaulting as Karl suggested (with pipes as that's much better way to use functions).

charts/posthog/values.yaml
fuziontech: perfect

charts/posthog/templates/clickhouse_cluster_role.yaml
```diff
@@ -124,5 +124,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: clickhouse-operator
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
```
tiina303, line 127:
```suggestion
  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }} 
```

charts/posthog/templates/clickhouse_config_map.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etc-clickhouse-operator-files
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
```
tiina303, line 6:
```suggestion
  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }} 
```

charts/posthog/templates/clickhouse_config_map.yaml
```diff
@@ -134,7 +134,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etc-clickhouse-operator-confd-files
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
```
tiina303, line 137:
```suggestion
  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
```

charts/posthog/templates/clickhouse_config_map.yaml
```diff
@@ -143,7 +143,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etc-clickhouse-operator-configd-files
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
```
tiina303, line 146:
```suggestion
  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
```

charts/posthog/templates/clickhouse_config_map.yaml
```diff
@@ -193,7 +193,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etc-clickhouse-operator-templatesd-files
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
```
tiina303, line 196:
```suggestion
  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
```

charts/posthog/templates/clickhouse_config_map.yaml
```diff
@@ -292,7 +292,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: etc-clickhouse-operator-usersd-files
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
```
tiina303, line 295:
```suggestion
  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
```

charts/posthog/templates/clickhouse_deployment.yaml
```diff
@@ -3,7 +3,7 @@ kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: clickhouse-operator
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
```
tiina303, line 6:
```suggestion
  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
```

charts/posthog/templates/clickhouse_service.yaml
```diff
@@ -8,7 +8,7 @@ kind: Service
 apiVersion: v1
 metadata:
   name: clickhouse-operator-metrics
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }}
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
```
tiina303, line 11:
```suggestion
  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
```

charts/posthog/templates/clickhouse_service_account.yaml
```diff
@@ -3,5 +3,5 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: clickhouse-operator
-  namespace: {{ .Values.clickhouseOperator.namespace | quote }} 
+  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.namespace }}
```
tiina303, line 6:
```suggestion
  namespace: {{ .Values.clickhouseOperator.namespace | default .Release.Namespace }}
```

charts/posthog/values.yaml
```diff
@@ -21,8 +21,8 @@ sentryDSN:
...
+  # -- Which namespace to install clickhouse operator to (defaults to namespace chart is installed to)
+  namespace: ""
```
tiina303, line 25: If it's empty str, then default function thinks that's a reasonable value only for no value will default be used and that's what we want.

```suggestion
  namespace:
```

tiina303 (changes requested):
I had a feeling the "" didn't work with default, so I tested it, found another typo. Namespace needs to be capitalized after release.

btw simple way to test this:
```
helm template posthog charts/posthog --set cloud="x" -n "blah" | grep "clickhouse_service_account" -A5
```
^ and look at what you see for the namespace

Issue #43, opened 2021-07-12
Generate clickhouse password dynamically


macobo:
## Is your feature request related to a problem?

Currently, the clickhouse password is static and publicly available in this repo

## Describe the solution you'd like

Generate the data dynamically via secrets.

## Describe alternatives you've considered

Make it a mandatory variable

## Additional context

See https://github.com/Altinity/clickhouse-operator/issues/700 for context on how to implement this.

tiina303:
Reprioritizing as we changed Clickhouse to be a Nodeport instead of a LB, i.e. it's not externally accessible anymore.

Note that https://github.com/PostHog/charts-clickhouse/pull/81 which (didn't solve the full problem) & ran into a problem if the `users.xml` didn't get properly recomputed (upgrade failed for whatever reason at inopportune time or clickhouse operator thought nothing needed to be updated) can be fixed by making the backup password (if an upgrade & we had a previous secret) to be the hardcoded one we have now and only generate new password dynamically for new installs.

See also https://github.com/PostHog/vpc/issues/165 for more context. Clickhouse operator password is stored in ConfigMap both as plaintext and sha, which I'm not sure if we can pull values from secrets for.

PR #81, opened 2021-07-30
Autogenerate clickhouse password for admin user

tiina303:
Part 1 for https://github.com/PostHog/charts-clickhouse/issues/43

Default user now doesn't have a password and is only accessible on localhost, which makes it nicer when we shell into the clickhouse pod, we can just do `clickhouse-client`. If someone is on localhost they have access to pretty much everything already, so I don't see that as a security risk, but I also couldn't figure out how to set the default user password from the autogenerated value (the obvious ways didn't merge into the `users.xml` file properly).

The admin user now get an autogenerated password that is used by all the other services, can be used to access clickhouse from the outside through the public ip, with password that can be looked up by
```console
kubectl get secret posthog -n posthog -o jsonpath="{.data['clickhouse-password']}" | base64 --decode
``` 

The only other change is removal of `host_regexp` from admin.network, but my previous testing in https://github.com/PostHog/vpc/issues/165 didn't seem to indicate any value for it.

Full diff of changes to `/var/lib/clickhouse/preprocessed_configs/users.xml`

```
diff --git a/users-before b/users-new-6
index bcb987b..57f79c3 100644
--- a/users-before
+++ b/users-new-6
@@ -8,7 +8,8 @@
        /etc/clickhouse-server/users.d/02-clickhouse-default-profile.xml
        /etc/clickhouse-server/users.d/03-database-ordinary.xml
        /etc/clickhouse-server/users.d/chop-generated-profiles.xml
-       /etc/clickhouse-server/users.d/chop-generated-users.xml      -->
+       /etc/clickhouse-server/users.d/chop-generated-users.xml
+       /etc/clickhouse-server/users.d/my_users.xml      -->
 
 <yandex>
     <!-- Profiles of settings. -->
@@ -93,7 +94,7 @@
                  Execute: PASSWORD=$(base64 < /dev/urandom | head -c8); echo "$PASSWORD"; echo -n "$PASSWORD" | sha1sum | tr -d '-' | xxd -r -p | sha1sum | tr -d '-'
                  In first line will be password and in second - corresponding double SHA1.
             -->
-            
+            <password/>
 
             <!-- List of networks with open access.
 
@@ -116,10 +117,11 @@
                  All results of DNS requests are cached till server restart.
             -->
             <networks>
-                <ip>0.0.0.0/0</ip>
+                <ip>::1</ip>
             
                 <host_regexp>(chi-posthog-[^.]+\d+-\d+|clickhouse\-posthog)\.posthog\.svc\.cluster\.local$</host_regexp>
                 
+                <ip>127.0.0.1</ip>
             </networks>
 
             <!-- Settings profile for user. -->
@@ -133,8 +135,6 @@
         
             
             
-            <password_sha256_hex>ea74fb006c130c4aebbb18b34fd65b0f787e31658ab3603a0712324a1966919c</password_sha256_hex>
-            
             
         </default>
     
@@ -149,17 +149,17 @@
             <quota>default</quota>
         </clickhouse_operator>
     
-        <admin>
-            <networks>
-                <host_regexp>(chi-posthog-[^.]+\d+-\d+|clickhouse\-posthog)\.posthog\.svc\.cluster\.local$</host_regexp>
-                <ip>0.0.0.0/0</ip>
-            </networks>
-            <password_sha256_hex>ea74fb006c130c4aebbb18b34fd65b0f787e31658ab3603a0712324a1966919c</password_sha256_hex>
-            <profile>default</profile>
-            <quota>default</quota>
-        </admin>
         
-    </users>
+    
+    <admin>
+      <password>aVtk6Sc3BrpqBkcX9WdYCqFztxEwnDcsSTMACxz2</password>
+      <profile>default</profile>
+      <quota>default</quota>
+      <networks>
+        <ip>0.0.0.0/0</ip>
+      </networks>
+    </admin>
+  </users>
 
     <!-- Quotas. -->
     <quotas>
@@ -208,4 +208,6 @@
     
 
     
+
+  
 </yandex>
\ No newline at end of file
```

#### Testing
1. Fresh install
1. 1-click DigitalOcean install + upgrade with these changes
1. Custom password in provided values + upgrade


fuziontech (approved):
This is really great work @tiina303 !

This improves the experience and security on every front. Well done!

I'm personally looking forward to not having to copy pasta a password when shelling into the clickhouse container.

charts/posthog/templates/secrets.yaml
```diff
@@ -16,6 +16,16 @@ data:
   {{- else }}
   posthog-secret: {{ randAlphaNum 40 | b64enc | quote -}}
   {{- end }}
+
+  {{- $clickhousePrevPassword := index ($previous.data | default dict) "clickhouse-password" }}
```
macobo, line 20: Code smell: the index call here different than on line 43 due to that checking if `$previous` exists first and here using a default. Consistency would be :kissing:

macobo (approved):
Great work! I assume the testing of this has already been done?

Note for local testing using k3s is awesome, really recommend it. 

Q: Why is this part 1? As I see it it solves the problem

tiina303:
> Note for local testing using k3s is awesome, really recommend it.

Thanks, will try that out, I've been using DigitalOcean, which is pretty great too.

> Q: Why is this part 1? As I see it it solves the problem

It doesn't handle all accounts that have passwords public (clickhouse_operator), which I mentioned here: https://github.com/PostHog/vpc/issues/165 so I'm planning to do that in a follow-up diff, i.e. part 2

macobo:
> mentioned in the end of my first comment

Wasn't sure if that was the plan or already done :)

tiina303:
Looks like k3s only likes Linux (https://rancher.com/docs/k3s/latest/en/installation/installation-requirements/#operating-systems) so I'll just use DigitalOcean

charts/posthog/templates/secrets.yaml
tiina303: Here I want it to be `nil` if we are doing an install (i.e. previous didn't exist) or if the "clickhouse-password" didn't exist (i.e. first upgrade after this change - to move folks away from using the hardcoded "a1..." if they haven't specified a password themselves). I could use an if/else block, but that seems worse. 
```
{{- if $previous }}
{{- $clickhousePrevPassword := index $previous.data "clickhouse-password" }}
{{- else }}
{{- $clickhousePrevPassword := nil }}
```

Similarly below for uniqName it's easier to use the previous check earlier as that applies to the whole block, I only ever want to fail there if we're upgrading (vs here I want to generate & change the password if previously we were using the hardcoded a1...). But if I were to try to get rid of the inconsistency I'd change it there, that however didn't seem to belong to this pr.

PR #83, opened 2021-08-03
Revert "Autogenerate clickhouse password for admin user"

tiina303:
Reverts PostHog/charts-clickhouse#81


fuziontech (approved):
🚢

tiina303:
If you're stuck in a weird state (ran an install/upgrade between earlier pr & this revert), you can look up the admin password with `kubectl exec chi-posthog-posthog-0-0-0 -- cat /var/lib/clickhouse/preprocessed_configs/users.xml` and add it to your values.yaml (`clickhouse.password`)

tiina303:
More context on the revert: discovered that `users.xml` doesn't always get regenerated when the password changes in secrets, not exactly sure yet what triggers the recompile, but anyway this can lead to `users.xml` having a different password (& that's the source of truth for ClickHouse) compared to what we have in secrets, which then causes migration (anything else trying to use that new password) to fail with this error
```
infi.clickhouse_orm.database.ServerError: admin: Authentication failed: password is incorrect or there is no user with such name  (version 21.6.5.37 (official build)) (516)
```

PR #85, opened 2021-08-03
Remove password and restrict to localhost for CH default user

tiina303:
The default settings for default user don't have a password & restrict to local host access only. They seem better compared to what we currently have (a hardcoded password that's in our public repo and access available from everywhere).

Checked the `users.xml` file to confirm no pass & only localhost access (`kubectl exec chi-posthog-posthog-0-0-0 -- cat /var/lib/clickhouse/preprocessed_configs/users.xml`). Also checked that if I shelled into the clickhouse pod I didn't need a password, i.e. `clickhouse-client` alone works.

We don't use the default account anywhere, so this should be safe.


fuziontech (approved):
💣

PR #86, opened 2021-08-03
Clickhouse from LB to a Nodeport

tiina303:
It's safer and cheaper. Part 2 of & closes https://github.com/PostHog/vpc/issues/166

Config info: ​​https://github.com/Altinity/clickhouse-operator/blob/master/docs/chi-examples/99-clickhouseinstallation-max.yaml

Tested with a helm upgrade.
Before
```
> kubectl get svc | grep "clickhouse"
clickhouse-operator-metrics                  ClusterIP      10.245.155.249   <none>          8888/TCP                        133m
clickhouse-posthog                           LoadBalancer   10.245.30.169    139.59.207.61   8123:32317/TCP,9000:32110/TCP   12m
```
After
```
> kubectl get svc | grep "clickhouse"
clickhouse-operator-metrics                  ClusterIP      10.245.155.249   <none>          8888/TCP                        138m
clickhouse-posthog                           NodePort       10.245.30.169    <none>          8123:30329/TCP,9000:31230/TCP   17m
```
Also nuked web task & it came back up fine & verified that the UI works displaying previously existing data.

### DIFF
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -36,6 +36,7 @@ spec:
   defaults:
     templates:
       dataVolumeClaimTemplate: data-volumeclaim-template
+      serviceTemplate: chi-service-template
   configuration:
     users:
       admin/password: {{ .Values.clickhouse.password }}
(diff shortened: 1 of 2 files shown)


fuziontech (approved):
This looks good, but is there any way we can make this an option using values to enable the LB? We do have one client who is loves having access to clickhouse on their on-prem deploy

tiina303:
> This looks good, but is there any way we can make this an option using values to enable the LB? We do have one client who is loves having access to clickhouse on their on-prem deploy

Yup, updated. But I'd still like the default to be NodePort (as I have configured it). For this one client if they should set their values file to have `clickhouseOperator.serviceType: LoadBalancer` before running the next upgrade then.

fuziontech:
Perfect

fuziontech:
test mention of PostHog/vpc#167

PR #87, opened 2021-08-04
Add plugins hpa template

jredl-va:
I noticed today that the plugins deployment wasn't scaling as expected when the hpa was enabled. It turns out the plugins hpa template didn't exist. 

After updating chart you can see the hpa has been created:

image

@tiina303 @fuziontech

### DIFF
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 3.2.0
+version: 3.3.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 3.3.0
+version: 3.2.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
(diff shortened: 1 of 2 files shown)


fuziontech (approved):
🎉🎉🎉

tiina303:
Thanks 😃  I'd love to hear how the autoscaling works for you, I'm not sure if all our services are primarily cpu bound.

jredl-va:
> Thanks 😃 I'd love to hear how the autoscaling is working for you. PS: I'm not sure if all our services are primarily cpu bound.

So far so good, still probably need to do some more tuning on what resources posthog is assigned. However here is the current footprint:

image

The two deployments that I've noticed cpu throttling (actually requiring scaling) are the web and plugins. The configuration we are currently running with. 

web:
```
web:
  # -- Web horizontal pod autoscaler settings
  hpa:
    # -- Boolean to create a HorizontalPodAutoscaler for web
    # -- This experimental
    enabled: true
    # -- CPU threshold percent for the web
    cputhreshold: 100
    # -- Min pods for the web
    minpods: 3
    # -- Max pods for the web
    maxpods: 10
  # -- Amount of web pods to run. Ignored if hpa is used
  replicacount: 1
  # -- Resource limits for web service. See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-requests-and-limits-of-pod-and-container for more
  resources:
     limits:
       cpu: 800m
       memory: 500Mi
     requests:
       cpu: 500m
       memory: 300Mi
```

Plugins:
```
plugins:
  ingestion:
    # -- Whether to enable plugin-server based ingestion
    enabled: true
  hpa:
    # -- Boolean to create a HorizontalPodAutoscaler for plugin server
    # -- This experimental, based on cpu util which is not necessarilly the bottleneck
    enabled: true
    cputhreshold: 100
    minpods: 2
    maxpods: 10
  env: []
  # -- How many replicas of plugin-server to run. Ignored if hpa is used
  replicacount: 1
  resources:
     limits:
       cpu: 4000m
       memory: 4000Mi
     requests:
       cpu: 2000m
       memory: 2500Mi
```

PR #88, opened 2021-08-04
Update redis configuration for plugins deployment

jredl-va:
The posthog plugins deployment is not configured correctly when using a cloud based redis instance (gcp managed redis for example). I've been manually updating the deployment to include these values as per the other deployments but figured I would take the toil out of my day by making a pull request. 

Note, that I also update the environment variable for the redis password as that doesn't look to align with the server configuration: https://github.com/PostHog/plugin-server/blob/0dcde9372d10b3c045c9d29464970edd8534b01b/src/config/config.ts#L44

@fuziontech @tiina303

### DIFF
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 3.2.0
+version: 3.3.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 3.3.0
+version: 3.2.0
 
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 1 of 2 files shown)


charts/posthog/templates/plugins-deployment.yaml
```diff
@@ -90,7 +90,7 @@ spec:
         - name: USING_PGBOUNCER
           value: 'true'
           {{- if or (.Values.redis.enabled) (.Values.redis.password) }}
-        - name: REDIS_PASSWORD
+        - name: POSTHOG_REDIS_PASSWORD
```
macobo, line 93: Great catch!

charts/posthog/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 3.2.0
+version: 3.3.0
```
macobo, line 18: Let's keep the previous version. We bump versions via merging PRs with appropriate labels (will handle it).

```suggestion
version: 3.2.0
```

macobo (commented):
This looks great, one minor issue around versioning.

Can you enable https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork - we can change it ourselves. :)

jredl-va:
Version number reverted!

macobo (approved):

PR #89, opened 2021-08-04
Add support for using a kubernetes secret with the gcp ingress

jredl-va:
Allow a tls secret to be specified rather than using the gke certificate manager: https://kubernetes.io/docs/concepts/services-networking/ingress/#tls

For all of our ingress objects within GKE we specify tls configurations via kubernetes secrets. When using a gcp ingress object and a user has specified the tls name this will provision the ingress object accordingly and suppress the managed certificate from being created. 

@fuziontech @tiina303


fuziontech (approved):
This is great! Thanks @jredl-va

PR #92, opened 2021-08-05
fix REDIS_URL redis password usage

tiina303:
Follow-up for https://github.com/PostHog/charts-clickhouse/pull/88

Also removed the unused secrets stuff

tested with `echo $REDIS_URL` in the shell in the plugins container.


fuziontech (approved):
lgtm

macobo (approved):

PR #93, opened 2021-08-06
Have migration task bypass pgbouncer and connect direct to pg for session stickiness

fuziontech:
This is in preparation for issue posthog/posthog#4941 where we want to create an index concurrently that will exceed our statement_timeout time.

### DIFF
--- a/charts/posthog/templates/migrate.job.yaml
+++ b/charts/posthog/templates/migrate.job.yaml
@@ -82,6 +82,7 @@ spec:
               name: {{ template "posthog.postgresql.secret" . }}
             {{- end }}
               key: {{ template "posthog.postgresql.secretKey" . }}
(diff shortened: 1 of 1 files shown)


charts/posthog/templates/migrate.job.yaml
```diff
@@ -83,11 +83,11 @@ spec:
             {{- end }}
               key: {{ template "posthog.postgresql.secretKey" . }}
         - name: POSTHOG_POSTGRES_HOST
```
macobo, line 85: Worth adding a comment here since we have similar values elsewhere.

```suggestion
        # Connect directly to postgres (without pgbouncer) to avoid time limits
        - name: POSTHOG_POSTGRES_HOST
```

tiina303 (approved):

PR #96, opened 2021-08-11
Remove posthog specific email values

tiina303:
Closes https://github.com/PostHog/posthog/issues/5464

To know if email is enabled we check if hostname is specified https://github.com/PostHog/posthog/blob/dae1d86edb305f95692fb23c52e47b17f39204a0/posthog/email.py#L36
our default values had it specified, so we thought email was set up for everyone.

#### Before
forgot password link gets to 502 or tells the email was sent if user didn't exist
Instance Status page was wrong
Screen Shot 2021-08-11 at 21 53 05

#### Now
Screen Shot 2021-08-11 at 21 59 41
Screen Shot 2021-08-11 at 22 00 08

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -215,11 +215,11 @@ plugins:
 # Outbound E-mails
 email:
   # -- Outbound email sender
-  from_email: hey@posthog.com
+  from_email:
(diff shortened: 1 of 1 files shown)


fuziontech (approved):
✅

PR #98, opened 2021-08-11
Split web to app and events pods

tiina303:
Closes https://github.com/PostHog/vpc/issues/79

The code is a lot of copy pasta, but I don't really know how to do this well as referencing things across files doesn't work. Could move things to _helpers, but maybe there's a better way?

Note that used the name `web` as instead of `app`, because then we continue to use the same deployment (i.e. when someone is upgrading they don't need to manually remove the old `web` deployment, making this a minor instead of a major upgrade).

Tested both that the UI works and posthog.capture works from python.


fuziontech:
I would split this further into two deployments and two separate services
- `app`
- `events`

There will be a lot of duplication, but that's not a huge issue right now.

charts/posthog/templates/web-deployment.yaml
```diff
@@ -210,3 +210,216 @@ spec:
           timeoutSeconds: {{ .Values.web.readinessProbe.timeoutSeconds }}
         resources:
 {{ toYaml .Values.web.resources | indent 12 }}
+---
```
fuziontech, line 213: Can you split this out into `events-deployment.yaml` This is getting a bit big

charts/posthog/templates/web-service.yaml
```diff
@@ -36,3 +36,42 @@ spec:
   loadBalancerSourceRanges:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+---
```
fuziontech, line 39: Can you split this out into `events-service.yaml` This is getting a bit big. This is also how we do every other service in the chart (split up into their own files)

fuziontech (approved):
It would be nice if we can split these out into their constituent files. After that 🚢 it

PR #103, opened 2021-08-24
nameOverride improvements

benbz:
e.g. setting in `values.yaml`:
```yaml
zookeeper:
  nameOverride: zookeeper
```
now results in a service `posthog-zookeeper` being created rather than `posthog-posthog-zookeeper`. The strings replace by `$name` in the `printf`s correspond to the default `nameOverride`s for postgresql, redis, & kakfa so will be a no-op if they haven't been customised


tiina303:
I noticed that there was some oddness in how we were using the Overrides in general, so I put up a pr to cover that + your changes here https://github.com/PostHog/charts-clickhouse/pull/107

PR #104, opened 2021-08-24
pgbouncer extension points

benbz:
e.g. setting in `values.yaml`:
```yaml
pgbouncer:
  env:
  - name: SERVER_TLS_SSLMODE
    value: require
  extraVolumeMounts:
  - name: config
    mountPath: /etc/pgbouncer
  extraVolumes:
  - name: config
    emptyDir: {}
```
now results in `server_tls_sslmode = require` in `pgbouncer.ini` and uses an `emptyDir` rather than writing files in the container. This is useful for OpenShift as the entrypoint script can't write files to `/etc/pgbouncer`. It is not sufficient on its own - a custom entrypoint script or `pgbouncer.ini` is also required so that `user = postgres` is replaced with `user =` as per https://github.com/edoburu/docker-pgbouncer/pull/23

### DIFF
--- a/charts/posthog/templates/pgbouncer-deployment.yaml
+++ b/charts/posthog/templates/pgbouncer-deployment.yaml
@@ -81,3 +81,9 @@ spec:
             port: 6543
           initialDelaySeconds: 15
           periodSeconds: 20
+      {{- if .Values.pgbouncer.extraVolumeMounts }}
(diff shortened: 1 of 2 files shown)


tiina303 (approved):
Thanks for improving our charts 😃

PR #106, opened 2021-08-24
Make the command to run for beat configurable

benbz:
e.g setting in `values.yaml`
```yaml
beat:
  command:
  - /bin/sh
  - -c
  - |
    celery -A posthog beat -S redbeat.RedBeatScheduler --pidfile=/tmp/celery.pid
```
allows the beat deployment to run on OpenShift as it has permissions to write the pidfile into `/tmp`.

I'm less sure about this PR. It might be more useful to change/make configurable the pidfile location in https://github.com/PostHog/posthog/blob/master/bin/docker-worker-beat instead.

### DIFF
--- a/charts/posthog/templates/beat-deployment.yaml
+++ b/charts/posthog/templates/beat-deployment.yaml
@@ -57,11 +57,7 @@ spec:
       - name: {{ .Chart.Name }}-beat
         image: {{ template "posthog.image.fullPath" . }}
         imagePullPolicy: {{ .Values.image.pullPolicy }}
-        command:
-          - /bin/sh
-          - -c
-          - |
-            ./bin/docker-worker-beat
+        command: {{- toYaml .Values.beat.command | nindent 8 }}
         ports:
         - containerPort: {{ .Values.service.internalPort }}
         env:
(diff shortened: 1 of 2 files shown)


tiina303:
> I'm less sure about this PR. It might be more useful to change/make configurable the pidfile location in https://github.com/PostHog/posthog/blob/master/bin/docker-worker-beat instead.

I agree. We might want to in the future change `bin/docker-worker-beat`, so users defining a full command like this would need to keep track of those changes and add them to their values and that's just not going to happen :D

I suspect the easiest way would be to get the pidfile location (or potentially even extra flags for the `celery beat` cmd) through an env variable to be used in https://github.com/PostHog/posthog/blob/master/bin/docker-worker-beat. Note that you can just define the appropriate sha for posthog version after the change has landed in the Posthog repo side, so no extra delay doing it this way. Open to your ideas.

guidoiaquinti:
We are going to deprecated the beat deployment via https://github.com/PostHog/charts-clickhouse/pull/184. @benbz let us know if we can close this.

benbz:
> We are going to deprecated the beat deployment via #184. @benbz let us know if we can close this.

Hi, sorry I got shifted onto other projects at work and didn't get a chance to circle back around to https://github.com/PostHog/charts-clickhouse/pull/106#issuecomment-906171226. Removing the beat deployment also solves my problem, so SGTM. Thanks!

PR #110, opened 2021-08-26
Remove docs from repo and link to posthog.com

tiina303:
We've decided docs should live in Posthog.com, removing from here so we don't have out of date info.


fuziontech (approved):
😢

PR #119, opened 2021-09-14
Set app version to 1.28.0 in chart.yaml

fuziontech:
for release of 1.28.0!


fuziontech:
@tiina303 sorry I should have called this out in your PR https://github.com/PostHog/charts-clickhouse/blob/main/charts/posthog/values.yaml#L12

tiina303:
where is that used, looks like we didn't update it for 1.27 either?

tiina303 (approved):

fuziontech:
it's used for listing information from `helm` against the repo. You can basically see what version of the chart will install what version of posthog just by using the helm cli. Pretty snazzy.

PR #121, opened 2021-09-14
Add kubeversion to helm install info

tiina303:
So we know what versions people are using which makes it easier to know what we need to keep supporting in the chart

Verified by deploying to my DigitalOcean instance:
Screen Shot 2021-09-15 at 01 54 51


fuziontech:
Unrelated question - Why is the Git SHA `Undefined`?

fuziontech (approved):
lgtm

tiina303:
> Why is the Git SHA Undefined?

The image wasn't built from the GH action I updated (it does work for latest packages), see https://posthog.slack.com/archives/C0113360FFV/p1631679679157500?thread_ts=1631634071.148600&cid=C0113360FFV

PR #123, opened 2021-09-15
SAML support

paolodamico:
Adds values require for SAML support. Please check out the docs for details on how this should work, https://posthog.com/docs/user-guides/sso#saml. I'm particularly not sure about the X509 certificate as it's a multi-line string


charts/posthog/values.yaml
```diff
@@ -408,11 +428,10 @@ clickhouse:
...
   # externalZookeeper:
-    # -- URL for zookeeper.
-    # servers:
-      # - host: posthog-posthog-zookeeper
-        # port: 2181
-
+  # -- URL for zookeeper.
+  # servers:
+  # - host: posthog-posthog-zookeeper
+  # port: 2181
```
tiina303, line 434: please revert this part

tiina303:
Maybe you could provide an example values entries that need to be added in the doc page too.

charts/posthog/templates/web-deployment.yaml
```diff
@@ -146,6 +146,28 @@ spec:
...
+          value: {{ default "" .Values.saml.attr_last_name | quote }}
+        - name: SAML_ATTR_EMAIL
+          value: {{ default "" .Values.saml.attr_email | quote }}
+        {{- if eq .Values.saml.enforced true }}
+          - name: SAML_ENFORCED
+            value: '1'
+        {{- end }}
+        {{- if eq .Values.saml.disabled true }}
+          - name: SAML_DISABLED
+            value: '1'
```
tiina303, line 169:
```suggestion
        - name: SAML_ENFORCED
          value: '1'
        {{- end }}
        {{- if eq .Values.saml.disabled true }}
        - name: SAML_DISABLED
          value: '1'
```

charts/posthog/values.yaml
```diff
@@ -242,6 +241,27 @@ email:
   # existingSecret: secret-name
   # existingSecretKey: smtp-password
 
+# SAML
+saml:
+  # -- Whether password-based login is disabled and users automatically redirected to SAML login. Requires SAML to be properly configured.
+  enforced: false
+  # -- Whether SAML should be completely disabled. If set at build time, this will also prevent SAML dependencies from being installed.
+  disabled: false
```
tiina303, line 249: shouldn't this be `true` by default?

tiina303: also the convention seems to be to rather have `saml.enabled: true/false`

paolodamico: Actually no, we only support disabling SAML in case you have an environment that doesn't support the needed deps (like Heroku) or don't want to install those deps for some reason, but it should be a rare case in this context.

As for `saml.disabled` it's following the convention of the env var that we read in the codebase (see `SAML_DISABLED` in https://posthog.com/docs/self-host/configure/environment-variables), but happy to change if you think it's clearer.

charts/posthog/values.yaml
paolodamico: done!

charts/posthog/values.yaml
```diff
@@ -242,6 +241,34 @@ email:
...
+  # -- Entity ID from your SAML IdP.
+  # entity_id: id-from-idp-5f9d4e-47ca-5080
+  entity_id:
+  # -- Assertion Consumer Service URL from your SAML IdP.
+  # acs_url: https://mysamlidp.com/saml2
```
paolodamico, line 254: Not sure if we need to escape this somehow, would appreciate some input

charts/posthog/values.yaml
```diff
@@ -242,6 +241,34 @@ email:
...
...50IDAeFw0yMTA4MTYyMTUyMzNaFw0yNjA4\nMTYyMTUyMzNaMEUxEDAOBgNVBAoMB1Bvc3RIb2cxFTATBgNVBAsMDE9uZUxvZ2luIElkUDEaMBgGA1UEAwwRT25lTG9naW4gQWNjb3VudCAwggEiMA0GCSqGSIb3DQEB\nAQUAA4IBDwAwggEKAoIBAQDEfUWFIU38ztF2EgijVsIbnlB8OIwkjZU8c34B9VwZ\nBQQUSxbrkuT9AX/5O27G04TBCHFZsXRId+ABSjVo8daCPu0d38Quo9KS3V3627Nw\nYcTYsje95lB02E/PgfiEQ6ZGCOV0P4xY9C99d26PoYTcoMT1S73jDDMOFtoD5WXG\nZsKqwBks1jbLkv6RYoFBlZX00aGzOXDzUXI59/0c15KR4EzgTad0t6CU7X0HZ2Qf\nxGUiRb7hDLvgSby0SzpQpYUyYDnN9aSNYzpu1hiyIqrhQ7kZNy7LyGBz0UIuIImF\npF6A3bzzrR4wdacFY9U0vmqFXXcepxuT5p2UyAxwbLeDAgMBAAGjgcMwgcAwDAYD\nVR0TAQH/BAIwADAdBgNVHQ4EFgQURLVVKanZPoXGEfYr1HmlaCEoD54wgYAGA1Ud\nIwR5MHeAFES1VSmp2T6FxhH2K9R5pWghKA+eoUmkRzBFMRAwDgYDVQQKDAdQb3N0\nSG9nMRUwEwYDVQQLDAxPbmVMb2dpbiBJZFAxGjAYBgNVBAMMEU9uZUxvZ2luIEFj\nY291bnQgghR2uIejyqYBTUjSDGwjtwVkeadrfzAOBgNVHQ8BAf8EBAMCB4AwDQYJ\nKoZIhvcNAQEFBQADggEBALP5lhlcV8avbnVnqO7PBtlS2mVOJ2B7obm50OaJCbRh\nt0I/dcNssWhT31/zmtNfKtrFicNImlKhdirApxpIp1WLEFY01a40GLmO6FG/WVvB\nEzwXonWP+cP8jYQnqZ15JkuHjP3DYJuOak2GqAJAfaGO67q6IkRZzRq6UwEUgNJD\nTlcsJAFaJDrcw07TY3mRFragdzGC7Xt/CM6r/0seY3+VBwMUMiJlvawcyQxap7om\nEdgmQkJA8Dk6f+geI+U7jV3orkPiofBJi9K6cp5Fd9usut8jwi3GYg2wExNGbhF4\nwlMD1LOhymQGBnTXPk+000nkBnYdqEnqXzVpDiCG1Pc=
```
paolodamico, line 257: Ideally we would display naturally, multiline instead of using the `\n` characters. Is there a way to support this?

fuziontech: an entire website dedicated to it! https://yaml-multiline.info/ :P

charts/posthog/values.yaml
fuziontech: wdym? What needs escaping?

charts/posthog/values.yaml
fuziontech: This seems reasonable enough to me with the comments

charts/posthog/values.yaml
paolodamico: Not sure if the `/` or `:` would need escaping for some reason ?

charts/posthog/values.yaml
paolodamico: thanks Jams!

paolodamico:
Alright @tiina303 @fuziontech feel free to merge if you think it's good to go

charts/posthog/values.yaml
tiina303: generally all strings in quotes is the suggestion from helm https://helm.sh/docs/chart_best_practices/values/#make-types-clear

paolodamico: Escaped

charts/posthog/values.yaml
```diff
@@ -381,7 +429,7 @@ kafka:
     # -- URL for zookeeper. Only set when internal zookeeper is disabled
     # -- IF using default clickhouse zookeeper use <deployment-name>-posthog-zookeeper
     servers:
-    - posthog-posthog-zookeeper:2181
+      - posthog-posthog-zookeeper:2181
```
tiina303, line 432:
```suggestion
        - posthog-posthog-zookeeper:2181
```

charts/posthog/values.yaml
```diff
@@ -411,8 +459,7 @@ clickhouse:
     # -- URL for zookeeper.
     # servers:
       # - host: posthog-posthog-zookeeper
-        # port: 2181
-
+      # - port: 2181
```
tiina303, line 462:
```suggestion
      #   port: 2181
```

charts/posthog/values.yaml
```diff
@@ -381,7 +429,7 @@ kafka:
     # -- URL for zookeeper. Only set when internal zookeeper is disabled
     # -- IF using default clickhouse zookeeper use <deployment-name>-posthog-zookeeper
     servers:
-    - posthog-posthog-zookeeper:2181
+        - posthog-posthog-zookeeper:2181
```
tiina303, line 432:
```suggestion
    - posthog-posthog-zookeeper:2181
```

tiina303 (approved):

PR #127, opened 2021-09-16
Bump sha to include fixes for instance status reports

fuziontech:
Include fixes in posthog/product-internal#141 that fix issues with instance status report payloads


paolodamico:
FYI we have a full patch release coming up https://github.com/PostHog/posthog/pull/6021

Issue #117, opened 2021-09-14
K8s v1.22 compatibility


tiina303:
we use v1beta1 quite a bit & it's not available going forward
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

tiina303:
We only need to change 
1. `ingress.yaml` for `networking.k8s.io/v1beta1`
2. clickhouse crds for `apiextensions.k8s.io/v1beta1`

Note that the frontend config is `networking.gke` not `k8s` which isn't getting deprecated yet and is still used in various docs e.g. https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#1.16-gke.3+_1

based on 
```
$ grep "/v1beta1" -r charts/posthog
charts/posthog/crds/clickhouse_installation_template.yaml:apiVersion: apiextensions.k8s.io/v1beta1
charts/posthog/crds/clickhouse_operator_configurations.yaml:apiVersion: apiextensions.k8s.io/v1beta1
charts/posthog/crds/clickhouse_installation.yaml:apiVersion: apiextensions.k8s.io/v1beta1
charts/posthog/templates/gce-frontend-config.yaml:apiVersion: networking.gke.io/v1beta1
charts/posthog/templates/ingress.yaml:apiVersion: networking.k8s.io/v1beta1
charts/posthog/templates/ingress.yaml:    networking.gke.io/v1beta1.FrontendConfig: "{{ .Release.Name }}-frontend-config"
```

tiina303:
TODO: add a 1.22 test to the ci as well

tiina303:
Based on the earlier grep (no more v1beta1's) the good news is we're going to be k8s 1.25 compatible after these changes as we don't have anything that will be deprecated there https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25

PR #128, opened 2021-09-21
Add k8s 1.22 test

tiina303:
To make sure we don't break it later. For #117


tiina303:
hmm... this is odd 1.22 was supposed to fail as I didn't run this on top of my changes for ClickHouse crds update that's odd, need to check what these tests do.

fuziontech (commented):
🚋 1.22

fuziontech (approved):

PR #130, opened 2021-09-23
Add resources to pgbouncer deployment

terrymunro:
Similar to the other HPA services.

Not sure if setting resources is the correct way to resolve the following issue, but it can help with:

> failed to get cpu utilization: missing request for cpu

In the Horizontal Pod Autoscaler


tiina303 (approved):
Yup this should help. Thanks for improving our charts.

PR #139, opened 2021-09-30
Add dispatch workflow for bumping the chart

tiina303:
We occasionally (too often) forget to add a bump label, this will add a workflow to the actions we can trigger instead of creating no-op PRs with the label.

It's a bunch of duplication with the bump-version file, but seemed complicated to make the triggers work on the same file.

Triggering will look similar to ee test deploy on vpc
Screen Shot 2021-09-30 at 19 36 20

### DIFF
--- a/.github/workflows/bump-version-workflow-dispatch.yaml
+++ b/.github/workflows/bump-version-workflow-dispatch.yaml
@@ -1,0 +1,80 @@
+name: Bump chart version & release workflow dispatch
+
+on: 
+    workflow_dispatch:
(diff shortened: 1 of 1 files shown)


tiina303:
Landing this because we forgot to bump the chart again with the last PR

tiina303:
You can trigger it from https://github.com/PostHog/charts-clickhouse/actions/workflows/bump-version-workflow-dispatch.yaml & you can see 1 run I did just to make sure it works correctly.

PR #141, opened 2021-10-01
Add value to ingress to allow IP access

macobo:
Currently some guides (https://posthog.com/docs/self-host/deploy/azure)
say that you can access the site via the IP when you really can not

I added an option to help with debugging.

Have _not_ tested this!


fuziontech:
Looks like the helm chart is broken for DO right now 404ing on the landing page. 👀 into it right now

tiina303:
I propose we instead just remove the host there, any reason not to?

macobo:
Feel free to take this PR over and do whatever is needed @tiina303 - intent was to kick something off due to support hero work making it clear this is confusing.

tiina303:
I merged the docs change in a way that doesn't relay on this (saw something weird trying to remove the host here & it's not high pri, so didn't investigate too much)

PR #142, opened 2021-10-01
nginx and GCP compatible ingress controller

fuziontech:
Nginx works with `Prefix` path definitions.
GCP requires `ImplementationSpecific` path definitions.
Switch between the two here based on Ingress that is configured

### DIFF
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -56,8 +56,13 @@ spec:
     - http:
   {{- end }}
         paths:
-          - pathType: ImplementationSpecific
+          {{- if (eq (include "ingress.type" .) "nginx") }}
+          - pathType: Prefix
+            path: "/"
+          {{- else }}
+          - pathType: ImplementationSpecific 
             path: "/*"
+          {{- end }}
             backend:
               {{- if semverCompare ">=1.19.0" .Capabilities.KubeVersion.Version }}
               service:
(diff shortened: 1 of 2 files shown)


charts/posthog/templates/ingress.yaml
```diff
@@ -56,8 +56,13 @@ spec:
     - http:
   {{- end }}
         paths:
+          {{- if (eq (include "ingress.type" .) "nginx") }}
```
tiina303, line 59: Let's swap the default - `Prefix` should work in most places as that's what k8s docs have & the regex thing is different & we can't easily make assumptions about how regex needs to be provided.

My comment from slack 
for nginx we can use the regex with ".*" style (https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/) if we add the annotation use-regex (https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#use-regex)
but GKE expects it to be "*" (https://cloud.google.com/kubernetes-engine/docs/concepts/ingress), so looks like we need to split it based on which ingress controller we have.

tiina303:
btw you can test this by running 
```
helm template posthog ../charts-clickhouse/charts/posthog --set cloud=gcp | grep "Ingress" -A50
```
compared to what you see on master & 
```
helm template posthog ../charts-clickhouse/charts/posthog --set cloud=do | grep "Ingress" -A50
```
compared to before we broke it for nginx

charts/posthog/templates/ingress.yaml
```diff
@@ -56,8 +56,13 @@ spec:
     - http:
   {{- end }}
         paths:
+          {{- if (ne (include "ingress.type" .) "clb") }}
```
tiina303, line 59: nit: 
instead of 
```
if not x:
  y 
else:
  z
```
I prefer
```
if x:
 z 
else:
  y
```

tiina303 (approved):
🚢

charts/posthog/templates/ingress.yaml
fuziontech: PR accepted ;)

tiina303: yup totally didn't mean for you to update again, that was more for future reference ;)

PR #144, opened 2021-10-01
Add unsafeHttpDirectAccess option as a quick way to allow ip access

tiina303:
Inspired from https://github.com/PostHog/charts-clickhouse/pull/141 but I wanted to make it even simpler & to have it work for all platforms. So we can just say that they should run a `helm upgrade` with `--set ingress.unsafeHttpDirectAccess=true` in the troubleshooting sections.

Tested in DigitalOcean that it still worked to have hostname set etc & that after the update with this flag I could access via IP directly (before couldn't)

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -338,6 +338,8 @@ ingress:
   annotations: {}
   # -- TLS secret to be used by the ingress.
   secretName:
+  # -- Troubleshooting tool to swich to http direct access to the IP. This overrides hostname, nginx.redirectToTLS, gcp.forceHttps, letsencrypt and web.secureCookies to false. Don't use in Production.
+  unsafeHttpDirectAccess: false
 
 postgresql:
   # -- Install postgres server on kubernetes (see below)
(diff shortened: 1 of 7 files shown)


charts/posthog/values.yaml
```diff
@@ -338,6 +338,8 @@ ingress:
   annotations: {}
   # -- TLS secret to be used by the ingress.
   secretName:
+  # -- Troubleshooting tool to swich to http direct access to the IP. This overrides hostname, nginx.redirectToTLS, gcp.forceHttps, letsencrypt and web.secureCookies to false. Don't use in Production.
```
macobo, line 341: In general from what I've seen helm charts try to avoid having a lot of "business logic" inside the charts and values, instead being a straight-forward mapping from values to a kubernetes manifest.

I like that. Having values which have so large of an impact is a trap, because while it looks appealing from the outside, it:
1. Will cause surprises for our users (e.g. some of their values will be overwritten)
2. It will cause bugs as we update ingestion in general - e.g. we won't remember to consider this case in particular

tiina303:
abandoning

PR #145, opened 2021-10-06
Update kafka storage values

yakkomajuri:
I don't know a thing about k8s/helm charts really so just putting this up as a way of bringing visibility to this discussion.

We just had a customer run out of space on the Kafka pod. This is highly problematic as it can lead us to drop events. It also crashes the plugin server.

We need a few things:

- A way to resize this easily (maybe a doc at first?)
- More storage allocated to this by deafult (the customer this happened to really does not have a lot of events at this stage)

@guidoiaquinti also pointed out we probably want to be on the safe side in terms of Kafka retention to pod storage ratio.

This PR requests more storage by default and makes the ratio 50%. Again, what do I know? Not a thing. So just giving this visibility.

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -417,10 +417,10 @@ kafka:
     # -- Enable persistence using PVC
     enabled: true
     # -- PVC Storage Request for kafka volume
-    size: 5Gi
+    size: 20Gi
   # -- A size-based retention policy for logs
-  # -- Should be less than kafka.persistence.size, ideally 70-80%
-  logRetentionBytes: _4_000_000_000
+  # -- Should be less than kafka.persistence.size, minimum 1GB
+  logRetentionBytes: _15_000_000_000
   # -- The minimum age of a log file to be eligible for deletion due to age
   logRetentionHours: 24
   zookeeper:


yakkomajuri:
Update: the approach for this customer was to delete the kafka pod (causing a loss of the events that were in Kafka) and run `helm upgrade` with the updated `values.yaml` file using the values in this PR actually (coincidentally)

charts/posthog/values.yaml
```diff
@@ -415,10 +415,10 @@ kafka:
...
     # -- PVC Storage Request for kafka volume
-    size: 5Gi
+    size: 20Gi
   # -- A size-based retention policy for logs
-  # -- Should be less than kafka.persistence.size, ideally 70-80%
-  logRetentionBytes: _4_000_000_000
+  # -- Should be less than kafka.persistence.size, ideally ~50%
+  logRetentionBytes: _10_000_000_000
   # -- The minimum age of a log file to be eligible for deletion due to age
   logRetentionHours: 24
```
guidoiaquinti, line 423: Should we change this as well? Didn't we recently change some topic settings to be in the order of weeks? How does this cluster setting plays with that? Is it a global override or a default for topics without an explicit retention policy?

guidoiaquinti:
I'm going to create a dedicated `[Discussion]` issue with more 🧠 stuff but the tl;dr; is that we need `helm` charts and documentation support for both:

* vertical service scale: this is usually an operation to "buy time" for an horizontal scale operation  
* horizontal service scale: this is usually an operation that involves dataset partitioning/sharding and a cluster rebalance operation

for all our service dependency in our stack: Kafka, ClickHouse and PostgreSQL.

Bonus point: `helm` charts should ship as well a basic monitoring/alerting stack.

charts/posthog/values.yaml
fuziontech: :+1: I would definitely bump this up to a week or so. The `logRetentionBytes` will be the constraining factor for larger customers and we should communicate that in the docs that if they want to have a buffer for downtime, increase the [PVC claim size](https://github.com/PostHog/charts-clickhouse/pull/145/files#diff-e5d75a211cc4472486a4b35840f9e60d261b574e167ab44decf6bf5e4b7c3606R418)

fuziontech (approved):
Nicely done! This is also something you can have the customer bump themselves by having them apply this to their overwriting `values.yaml` file. There is a balance here between the cost of a small deployment and scalability. Here I think it's worth bumping the PVC from 5gb to 20gb, but there are some installs out there were every $1 we bump the less likely they are to keep the stack up.

charts/posthog/values.yaml
tiina303: It's the minimum retention before allowing someone to be deleted, which means that if we set this to 1 week, then anything that's only 6 days old can't be deleted even if the disk is completely full

tiina303 (changes requested):
This would trigger everyone's upgrades to fail because, with messages like
```
Error: UPGRADE FAILED: cannot patch "posthog-posthog-kafka" with kind StatefulSet: StatefulSet.apps "posthog-posthog-kafka" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```

If we want to do this we minimally need to write up the major upgrade notes, that said we don't have a good story here yet, so let's not land this. If we want customers to asap get better configs lets add the size change to our docs now while we figure out how to do the chart upgrade reasonably

charts/posthog/values.yaml
tiina303: me too & everyone else 😅  Related comment thread with more info https://github.com/PostHog/vpc/pull/110#discussion_r642531233

tiina303:
Tested on DigitalOcean by following @guidoiaquinti 'd guide which I linked to in Major upgrade notes here: https://github.com/PostHog/posthog.com/pull/2411 
I was thinking it would be nice for us to verify this on AWS as well or other platforms, where we might not have expandable volumes by default, but there's the alternative to set the values to old values + we can just deal with any support questions as/if they come up.

charts/posthog/values.yaml
tiina303: see https://github.com/PostHog/charts-clickhouse/issues/166

charts/posthog/values.yaml
```diff
@@ -417,10 +417,10 @@ kafka:
     # -- Enable persistence using PVC
     enabled: true
     # -- PVC Storage Request for kafka volume
-    size: 5Gi
+    size: 20Gi
```
guidoiaquinti, line 420: as a post merge follow up, we can probably remove all the `kafka.size` overrides we added in the docs as initial mitigation

tiina303: yup, that's the plan & part of https://github.com/PostHog/posthog.com/pull/2411

tiina303 (approved):

PR #179, opened 2021-11-03
Remove all the (possible) Helm annotations

guidoiaquinti:
## Description
1️⃣ Remove all the (possible) Helm annotations.
2️⃣ Add new `initContainer` definition to all the PostHog Deployments.
3️⃣ Group and cleanup `metadata.labels`.

## Why

1️⃣ There isn't a reason to have those hooks and they are breaking our chart in different ways (e.g. breaking a clean run of install + uninstall as due to those hooks we break some dependencies before removal, run updates without downtime, etc...). We are keeping the bare minimum annotations we need (for now 1).

2️⃣ Now that we don't enforce anymore a deployment order via Helm, let's introduce an `initContainer` definition to avoid Deployment pods crashing and restarting till all the external dependencies are not ready.

3️⃣ I took the opportunity to cleanup and standardise a bit this part.  

Release notes are available here https://github.com/PostHog/posthog.com/pull/2795

## Type of change

## How has this been tested?
1. CI is ✅ (except for the upgrade test as we need to perform some manual steps before running the upgrade)
1. CI for e2e test is also ✅ (I've triggered [DO](https://github.com/PostHog/charts-clickhouse/runs/4883398969?check_suite_focus=true), [GCP](https://github.com/PostHog/charts-clickhouse/runs/4883394313?check_suite_focus=true) and [AWS](https://github.com/PostHog/charts-clickhouse/runs/4883385445?check_suite_focus=true) tests manually) 
2. I've manually tested the upgrade path locally: Helm install from `main` -> manual upgrade steps -> Helm install from this branch -> ✅ 

## Checklist

### DIFF
--- a/.github/workflows/test-amazon-web-services-install.yaml
+++ b/.github/workflows/test-amazon-web-services-install.yaml
@@ -100,13 +100,26 @@ jobs:
     - name: Workaround - wait for all the k8s resources to be ready
       timeout-minutes: 15
       run: |
-        echo "Waiting for pods to be ready..."
-        while ! kubectl wait --for=condition=Ready pods --timeout=60s --all -n posthog > /dev/null 2>&1
+        echo "Waiting for jobs to be ready..."
+        while ! kubectl -n posthog wait --for=condition=Complete jobs --all --timeout 10s > /dev/null 2>&1
         do
-          echo "  pods are not yet ready"
+          sleep 1
         done
-        echo "All pods are now ready!"
+        echo "All jobs are now ready!"
 
+        echo "Waiting for all deployments to be ready..."
+        for deployment in $(kubectl -n posthog get deployments --no-headers -o custom-columns=NAME:.metadata.name)
+        do
+          while ! kubectl -n posthog rollout status deployment "$deployment" > /dev/null 2>&1
+          do
+            sleep 1
+          done
+        done
+        echo "All deployments are now ready!"
+
+    - name: Workaround - wait for the AWS load balancer to be ready
+      timeout-minutes: 15
+      run: |
         echo "Waiting for the AWS Load Balancer to be ready..."
         load_balancer_external_hostname=""
         while [ -z "$load_balancer_external_hostname" ];
--- a/charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
+++ b/charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
@@ -1,0 +1,51 @@
+{{/* Common initContainers-wait-for-migrations definition */}}
+{{- define "_snippet-initContainers-wait-for-migrations" }}
+- name: wait-for-migrations
+  image: {{ include "posthog.image.fullPath" . }}
+  command:
+    - /bin/sh
+    - -c
+    - >
+        until (python manage.py migrate --check);
+        do
+            echo "Waiting for database migrations to be completed"; sleep 1;
+        done
+  env:
+
+  # PostgreSQL configuration
+  - name: POSTHOG_POSTGRES_HOST
+    value: {{ template "posthog.pgbouncer.host" . }}
+  - name: POSTHOG_POSTGRES_PORT
+    value: {{ include "posthog.pgbouncer.port" . | quote }}
+  - name: POSTHOG_DB_NAME
+    value: {{ default "posthog" .Values.postgresql.postgresqlDatabase | quote }}
+  - name: POSTHOG_DB_USER
+    value: {{ default "posthog" .Values.postgresql.postgresqlUsername | quote }}
+  - name: POSTHOG_DB_PASSWORD
+    valueFrom:
+      secretKeyRef:
+      {{- if .Values.postgresql.existingSecret }}
+        name: {{ .Values.postgresql.existingSecret }}
+      {{- else }}
+        name: {{ template "posthog.postgresql.secret" . }}
+      {{- end }}
+        key: {{ template "posthog.postgresql.secretKey" . }}
+
+  # Redis configuration
+  - name: POSTHOG_REDIS_HOST
+    value: {{ template "posthog.redis.host" . }}
+  - name: POSTHOG_REDIS_PORT
+    value: {{ include "posthog.redis.port" . | quote }}
+  {{- if or (.Values.redis.enabled) (.Values.redis.password) }}
+  - name: POSTHOG_REDIS_PASSWORD
+    value: {{ .Values.redis.password | quote }}
+  {{- end }}
+
+  # Django specific settings
+  - name: SECRET_KEY
+    valueFrom:
+      secretKeyRef:
+        name: {{ template "posthog.fullname" . }}
+        key: posthog-secret
+
+{{- end }}
--- a/charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
+++ b/charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
@@ -1,0 +1,41 @@
+{{/* Common initContainers-wait-for-service-dependencies definition */}}
+{{- define "_snippet-initContainers-wait-for-service-dependencies" }}
+- name: wait-for-service-dependencies
+  image: busybox:1.34
+  command:
+    - /bin/sh
+    - -c
+    - >
+        {{ if .Values.clickhouse.enabled }}
+        until (nc -vz "{{ include "posthog.clickhouse.fullname" . }}.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local" 8123);
+        do
+            echo "waiting for ClickHouse"; sleep 1;
+        done
+        {{ end }}
+
+        until (nc -vz "{{ include "posthog.pgbouncer.host" . }}.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local" {{ include "posthog.pgbouncer.port" . }});
+        do
+            echo "waiting for PgBouncer"; sleep 1;
+        done
+
+        {{ if .Values.postgresql.enabled }}
+        until (nc -vz "{{ include "posthog.postgresql.host" . }}.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local" {{ include "posthog.postgresql.port" . }});
+        do
+            echo "waiting for PostgreSQL"; sleep 1;
+        done
+        {{ end }}
+
+        {{ if .Values.redis.enabled }}
+        until (nc -vz "{{ include "posthog.redis.host" . }}.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local" {{ include "posthog.redis.port" . }});
+        do
+            echo "waiting for Redis"; sleep 1;
+        done
+        {{ end }}
+
+        {{ if .Values.kafka.enabled }}
+        until (nc -vz "{{ include "posthog.kafka.host" . }}.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local" 9092);
+        do
+            echo "waiting for Kafka"; sleep 1;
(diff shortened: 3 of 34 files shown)


tiina303:
I think we might want hooks or another way to define the order, specifically we should (based on my limited understanding) start/upgrade stuff in this order:
1. DBs (zk, ClickHouse, Kafka, Redis)
2. Migration (need DBs to be up)
3. Services (web, events, worker) as they could depend on the migrations - e.g. a column doesn't exist yet as we haven't completed that migration, but we're already trying to process an event that trying to write to that column

Additionally I believe currently the migration task keeps crash looping in the beginning, which can be confusing to the users, it would be great if we could start that up after DBs are up fully.

guidoiaquinti:
Specifying a start-up order is an anti-pattern in Kubernetes. The idea is to make service resilient to dependency/service failures and leverage liveness and readiness checks to make sure everything is working. The only legit use I see for ordering is with _Custom Defined Resources_.  Here’s a nice and more detailed explanation I found online:

> Applications shouldn’t crash because a dependency isn’t ready. In traditional development, there is a specific order to the startup and stop tasks when bringing up applications. It’s important not to bring this mindset into container orchestration. With Kubernetes, Docker, etc., these components start concurrently, making it impossible to define a startup order. Even when the application is up and running, its dependencies could fail or be migrated, leading to further issues. The Kubernetes reality is also riddled with myriad points of potential communication failures where dependencies can’t be reached, during which a pod might crash or a service might become unavailable. Network latency, like a weak signal or interrupted network connection, is a common culprit for communication failure.
>
> ref: https://betterprogramming.pub/10-antipatterns-for-kubernetes-deployments-e97ce1199f2d#5ae4

For our specific case, Helm collects all the resources in a Chart and its dependencies, groups them by resource type, and then installs them in [this order](https://github.com/helm/helm/blob/v3.7.0/pkg/releaseutil/kind_sorter.go#L28-L66). Similar thing happens during the [uninstall](https://github.com/helm/helm/blob/v3.7.0/pkg/releaseutil/kind_sorter.go#L68-L106). 

By specifying a custom order via Helm hooks, we are introducing additional complexity that will probably lead to problems. For example, an install/uninstall operation of our chart will currently not succeed because of the custom ordering hooks we added.

tiina303:
thanks, so ignore the dbs being up before migrations part, but migrations need to be ran before services are installed/updated which is what's currently happening (via the hook weight) and we do need that ordering. To bring a specific example: Yakko recently added a version column via a migration & started using it in code afterwards, the code that tries to use the version column would crash (likely in a way we'd lose data) if the migration wasn't done beforehand.

I think we can remove the hook from migration job if we want, and keep it for events/web only.

guidoiaquinti:
> thanks, so ignore the dbs being up before migrations part, but migrations need to be ran before services are installed/updated which is what's currently happening (via the hook weight) and we do need that ordering. To bring a specific example: Yakko recently added a version column via a migration & started using it in code afterwards, the code that tries to use the version column would crash (likely in a way we'd lose data) if the migration wasn't done beforehand.

If we have a bug in the application that can lead to data loss when the app crashes, that’s another story; I think we should treat it as a P0 bug and this shouldn’t have anything to do with deployments. @yakkomajuri @tiina303 can you expand on this bug? Do we have it tracked somewhere?

I don’t see why we should enforce any sort of ordering, especially looking at what we just discussed. The application pod(s) will fail the readiness/liveness check(s) and will get restarted until the migration job succeeds.

guidoiaquinti:
As pointed by Karl, here’s some [additional context](https://github.com/PostHog/charts-clickhouse/pull/29) regarding why we added those hooks in first place. 

If install, upgrade and uninstall works on all platforms I’m inclined to move on once we clear out what we are discussing above (app bug).

tiina303:
With regards to migrations, I'm happy with any solution that assures we're not running web/plugins code before migrations are done (be it with deployments ordering (as is happening now) or the pods somehow crash looping until migrations are done). Potentially this is already the case that we have some checks around this in place, but we should verify that if we are removing the deployments way of assuring migrations are done before.

Note that sometimes in development we might want migrations to be in a weird state (when running PostHog plugins/backend locally).

guidoiaquinti:
@tiina303 with https://github.com/PostHog/posthog/pull/7823 I think we addressed your concern. I'm going to merge this PR once it's working fine and after we cut the next posthog release.

fuziontech:
I just tested this locally after testing the upgrade commands. Works like a charm. ✅

guidoiaquinti:
@fuziontech before merging this we need to:

1. wait for the next PostHog release in order to be sure https://github.com/PostHog/posthog/pull/7823 is included. 
1. wait till `ttlSecondsAfterFinished` is supported in all k8s versions we support (it is an API only available for v >= v1.21 and that's why I've removed the e2e testing for v1.20).

tiina303:
> wait for the next PostHog release in order to be sure do not start worker, plugins, or events if migrations not complete posthog#7823 is included.

People could still be running old posthog versions by pinning image.tag/sha, this shouldn't be a blocker, but lets just say in major upgrade notes that you can't run Posthog versions earlier than X anymore, on that note would it make sense for the chart to validate that similarly to k8s etc versions?

guidoiaquinti:
> People could still be running old posthog versions by pinning image.tag/sha, this shouldn't be a blocker, but lets just say in major upgrade notes that you can't run Posthog versions earlier than X anymore, on that note would it make sense for the chart to validate that similarly to k8s etc versions?

👍 to add an item to the release note.

guidoiaquinti:
I need to test this more but I think we can simply patch the deployments instead of deleting them:

Example:
```
kubectl label deployment posthog-events -n posthog app.kubernetes.io/managed-by=Helm
kubectl label deployment posthog-events -n posthog meta.helm.sh/release-name=posthog
kubectl label deployment posthog-events -n posthog meta.helm.sh/release-namespace=posthog
```

guidoiaquinti:
> > wait for the next PostHog release in order to be sure do not start worker, plugins, or events if migrations not complete posthog#7823 is included.
> 
> People could still be running old posthog versions by pinning image.tag/sha, this shouldn't be a blocker, but lets just say in major upgrade notes that you can't run Posthog versions earlier than X anymore, on that note would it make sense for the chart to validate that similarly to k8s etc versions?

Fixed by adding an `initContainer` definition for that (see above)

.github/workflows/test-amazon-web-services-install.yaml
```diff
@@ -100,13 +100,26 @@ jobs:
     - name: Workaround - wait for all the k8s resources to be ready
       timeout-minutes: 15
       run: |
-        echo "Waiting for pods to be ready..."
-        while ! kubectl wait --for=condition=Ready pods --timeout=60s --all -n posthog > /dev/null 2>&1
+        echo "Waiting for jobs to be ready..."
```
hazzadous, line 103: non-blocking

Worth pushing into a script, so it's easier to test locally, and also shared between aws and do?

charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
```diff
@@ -0,0 +1,51 @@
+{{/* Common initContainers-wait-for-migrations definition */}}
+{{- define "_snippet-initContainers-wait-for-migrations" }}
+- name: wait-for-migrations
+  image: {{ include "posthog.image.fullPath" . }}
+  command:
+    - /bin/sh
+    - -c
+    - >
+        until (python manage.py migrate --check);
```
hazzadous, line 9: I can't remember what the outcome was here, should we also be waiting on migration_clickhouse and anything else that might need to be run?

charts/posthog/templates/migrate.job.yaml
```diff
@@ -2,20 +2,13 @@
...
-    "helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"
-    "helm.sh/hook-weight": "0"
+  name: {{ printf "%s-migrate-%s" .Release.Name (now | date "2006-01-02-15-04-05") }}
+  labels: {{- include "_snippet-metadata-labels-common" . | nindent 4 }}
+  annotations: {{- include "_snippet-metadata-annotations-common" . | nindent 4 }}
 spec:
   template:
     metadata:
-      name: "{{ .Release.Name }}-migrate"
+      name: {{ printf "%s-migrate-%s" .Release.Name (now | date "2006-01-02-15-04-05") }}
```
hazzadous, line 11: What's the 2006 date for?

guidoiaquinti: it's only the date format (from go templates)

charts/posthog/templates/events-deployment.yaml
```diff
@@ -3,17 +3,8 @@ apiVersion: apps/v1
...
-    release: "{{ .Release.Name }}"
-    heritage: "{{ .Release.Service }}"
-  annotations:
-    # This is what defines this resource as a hook. Without this line, the
-    # job is considered part of the release.
-    "helm.sh/hook": "post-install,post-upgrade"
-    "helm.sh/resource-policy": "keep"
-    "helm.sh/hook-weight": "1"
+  labels: {{- include "_snippet-metadata-labels-common" . | nindent 4 }}
+  annotations: {{- include "_snippet-metadata-annotations-common" . | nindent 4 }}
```
tiina303, line 7: ❤️  so much cleaner & easier to change together in the future

charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
```diff
@@ -0,0 +1,51 @@
...
+        do
+            echo "Waiting for database migrations to be completed"; sleep 1;
+        done
+  env:
+
+  # PostgreSQL configuration
```
tiina303, line 15: this isn't super intuitive that it's in a file named `wait-for-migrations`

charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
```diff
@@ -0,0 +1,41 @@
+{{/* Common initContainers-wait-for-service-dependencies definition */}}
+{{- define "_snippet-initContainers-wait-for-service-dependencies" }}
+- name: wait-for-service-dependencies
+  image: busybox:1.34
+  command:
+    - /bin/sh
+    - -c
+    - >
+        {{ if .Values.clickhouse.enabled }}
```
tiina303, line 9: Maybe order them based on how fast they get ready usually (ClickHouse I suspect is last), so it feels like progress is happening and we get more detailed info from these logs.

tiina303 (approved):
The `initcontainers` approach looks great, we'd see the `echo` statements on the cli as the  install/upgrade is going, right?

charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
guidoiaquinti: PostgreSQL env variables?

charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
guidoiaquinti: We don't control how resources are created so I don't think there's a real gain of reordering those. Also, if a service is up and running the `nc` calls takes milliseconds

guidoiaquinti:
> we'd see the echo statements on the cli as the install/upgrade is going, right?
Nope pod logs only

charts/posthog/templates/web-hpa.yaml
```diff
@@ -2,12 +2,7 @@
 apiVersion: autoscaling/v1
 kind: HorizontalPodAutoscaler
 metadata:
-  name: {{ template "posthog.fullname" . }}-web
```
Legion2, line 5: the `name` attribute was removed, which causes erros on upgrade if hpa is enabaled

guidoiaquinti: It should be fixed now @Legion2 , sorry for the issue!

PR #187, opened 2021-11-15
[DigitalOcean 1 click] Use secure settings by default

guidoiaquinti:
## Description
Align _DigitalOcean 1 click setup_ to the _manual DO setup_.

The DO 1 click-setup is currently a two step process as described in our [docs](https://posthog.com/docs/self-host/deploy/digital-ocean). This is due to a limitation in DO marketplace API as it is not yet possible to provide parameters to DigitalOcean, so we need a post-install step to enable TLS.

This PR aligns `values.yaml` files between the two setups, after this change the only diff will be `ingress.hostname: <your-hostname>` missing (but that's expected as we can't use custom params in the 1 click install).

## Type of change

## How has this been tested?

## Checklist

### DIFF
--- a/do_1_click_values.yaml
+++ b/do_1_click_values.yaml
@@ -4,6 +4,5 @@ deploymentType: "DigitalOcean 1-click"
 ingress:
   nginx:
     enabled: true
-    redirectToTLS: false
-web:
-  secureCookies: false
+cert-manager:
+  enabled: true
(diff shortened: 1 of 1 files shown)


do_1_click_values.yaml
```diff
@@ -4,7 +4,5 @@ deploymentType: "DigitalOcean 1-click"
 ingress:
   nginx:
     enabled: true
-    redirectToTLS: false
-  letsencrypt: false
-web:
-  secureCookies: false
```
tiina303, line 10: we need these to be able to access without TLS (at least in the past)

guidoiaquinti:
As discussed on Slack, we still want to keep the non-TLS version. Closing

Issue #183, opened 2021-11-04
e2e chart testing


tiina303:
## Is your feature request related to a problem?

Unit testing is great and we're already working on it. Adding full e2e tests would give us even more confidence and to be able to move fast without breaking things.

## Describe the solution you'd like

We create e2e tests for all our platforms (DO, AWS, Azure, GCP).
We'll have a cluster we continuously use only for these tests, we'll run max 1 concurrently per cluster. This means these won't be ran on each PR most likely but rather on master in PostHog repo, which helps also catch when PostHog app version breaks data ingestion completely. We'll use this for automated release branch testing too.

The test will:
1. send some data & check it's there
2. try to upgrade the existing deployment and check that the data is still there
3. nuke the cluster & try a new install

## Describe alternatives you've considered

continue to manually test things

## Additional context

We already have https://github.com/PostHog/vpc/blob/main/client_values/posthog/values.yaml & https://github.com/PostHog/posthog/actions/workflows/vpc-deploy.yml

macobo:
This doesn't invalidate this ticket, but there's already a test deployment running on every PR: https://github.com/PostHog/charts-clickhouse/blob/main/.github/workflows/test-helm-chart.yaml

The suite is currently failing for unknown reasons. 

Before getting into setting up tests in more advanced configurations, it is worthwhile to get this in order as well.

tiina303:
Among others we should set-up e2e tests against https://github.com/PostHog/charts-clickhouse/blob/main/do_1_click_values.yaml (Do 1-click values file)

guidoiaquinti:
> This doesn't invalidate this ticket, but there's already a test deployment running on every PR: https://github.com/PostHog/charts-clickhouse/blob/main/.github/workflows/test-helm-chart.yaml
> 
> The suite is currently failing for unknown reasons.
> 
> Before getting into setting up tests in more advanced configurations, it is worthwhile to get this in order as well and use the same test suite for checking whether ingestion works as in there.

I did a first pass with https://github.com/PostHog/charts-clickhouse/pull/189. More will follow.

yakkomajuri:
Done given we deprioritized Azure

PR #180, opened 2021-11-03
CI - add 'uninstall' integration test

guidoiaquinti:
## Description
Add an e2e test to install and uninstall PostHog on a local k3s cluster.

Notes:
  - resources installed outside 'helm' (example: ClickHouse pods) will need to be removed manually
  - this test only verify if 'helm uninstall' works

## Type of change
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## How has this been tested?
CI should be ✅ 

## Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #188, opened 2021-11-16
e2e - DigitalOcean 1-click (install)

guidoiaquinti:
## Description
Add a new test to verify the DigitalOcean 1-click install. This workflow creates a new K8S cluster in DigitalOcean for each push this repo, waits until resources are ready to then reprovision it when the test is completed. 

I've opted for this solution instead of using a single cluster where to install/update/uninstall (as requested in https://github.com/PostHog/charts-clickhouse/issues/183) as there's no way atm to run a clean PostHog uninstall. Starting from scratch all the times is a bit more deterministic (from a cost perspective I don't think this should be an issue, the cluster costs $45/month and the test suite runs in ~ 10 min post cluster creation -> 45 / 43800 * 10 = ~ $0.01 per test).

This e2e test is not fully complete as we should also make sure:

1. all the expected k8s resources are in the state we expect (goal is to have a test suite using something like `vapor-ware/kubetest`)
2. ingestion test runs successfully (we have those but we need to adjust them a bit in order to do ingestion & checks via the LB -> see https://github.com/PostHog/charts-clickhouse/pull/189)
3. TLS setup is working (step 2 of "securing DO installation" guide. I need to figure out the domain management part yet)

I’m going to take ☝️ as separate PRs as they are pretty big and not exactly part of this task (they will be also shared across the other e2e tests).

## Type of change

## How has this been tested?
See CI

## Checklist

### DIFF
--- a/.github/workflows/test-install-digital-ocean.yaml
+++ b/.github/workflows/test-install-digital-ocean.yaml
@@ -26,6 +26,7 @@ jobs:
         doctl k8s clusters create \
           ${{ steps.vars.outputs.k8s_cluster_name }} \
           --region fra1 \
+          --version 1.21.5-do.0 \
           --tag="provisioned_by:github_action" \
           --size s-2vcpu-2gb \
           --count 3 \
--- a/.github/workflows/test-install-digital-ocean.yaml
+++ b/.github/workflows/test-install-digital-ocean.yaml
@@ -1,0 +1,82 @@
+name: e2e - DigitalOcean 1-click (install)
+
+on: [push, pull_request]
+
+jobs:
+  do-install:
+    runs-on: ubuntu-latest
+    if: github.repository == 'PostHog/charts-clickhouse'
+    steps:
+
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Install doctl
+      uses: digitalocean/action-doctl@v2
+      with:
+        token: ${{ secrets.DIGITALOCEAN_ACCESS_TOKEN }}
+
+    - name: Declare variables that we can share across steps
+      id: vars
+      run: |
+        echo "::set-output name=k8s_cluster_name::helm-e2e-testing-do-1-click-$(git rev-parse --short HEAD)"
+
+    - name: Deploy a new k8s cluster
+      run: |
+        doctl k8s clusters create \
+          ${{ steps.vars.outputs.k8s_cluster_name }} \
+          --region fra1 \
+          --tag="provisioned_by:github_action" \
+          --size s-2vcpu-2gb \
+          --count 3 \
+          --wait \
+          --1-clicks \
+          posthog
+
+    # Wait for the DigitalOcean marketplace to complete the installation
+    #
+    # Despite the --wait flag used in the command above
+    # there is no guarantee the PostHog app will be deployed
+    # when the command returns. With this workaround we wait
+    # until the Kubernetes keyspace is created.
+    #
+    # ref: https://github.com/digitalocean/doctl/issues/1063
+    #
+    # Why can't we directly use the 'action-k8s-await-workloads' step below?
+    # Because it's not working for this use case
+    #
+    # ref: https://github.com/jupyterhub/action-k8s-await-workloads/issues/38
+    #
+    - name: Workaround - wait until the 'posthog' pods are ready
+      timeout-minutes: 15
+      run: |
+        echo "Waiting for pods to be ready..."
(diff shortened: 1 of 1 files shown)


.github/workflows/test-install-digital-ocean.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - name: Declare variables that we can share across steps
+      id: vars
+      run: |
+        echo "::set-output name=k8s_cluster_name::helm-e2e-testing-do-1-click-$(git rev-parse --short HEAD)"
+
+    - name: Deploy a new k8s cluster
+      run: |
+        doctl k8s clusters create \
+          ${{ steps.vars.outputs.k8s_cluster_name }} \
+          --region fra1 \
```
guidoiaquinti, line 28: Looks like one of the fastest regions I've tested

.github/workflows/test-install-digital-ocean.yaml
```diff
@@ -0,0 +1,82 @@
...
+          --tag="provisioned_by:github_action" \
+          --size s-2vcpu-2gb \
+          --count 3 \
+          --wait \
+          --1-clicks \
```
guidoiaquinti, line 33: Directly test the 1-click / marketplace install

.github/workflows/test-install-digital-ocean.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+    - name: Deploy a new k8s cluster
+      run: |
+        doctl k8s clusters create \
+          ${{ steps.vars.outputs.k8s_cluster_name }} \
+          --region fra1 \
+          --version 1.21.5-do.0 \
+          --tag="provisioned_by:github_action" \
+          --size s-2vcpu-2gb \
+          --count 3 \
```
tiina303, line 32: Our docs say min req is count 2, so lets use 2 here as well

.github/workflows/test-install-digital-ocean.yaml
```diff
@@ -0,0 +1,83 @@
...
+    #
+    # ref: https://github.com/digitalocean/doctl/issues/1063
+    #
+    # Why can't we directly use the 'action-k8s-await-workloads' step below?
+    # Because it's not working for this use case
+    #
+    # ref: https://github.com/jupyterhub/action-k8s-await-workloads/issues/38
+    #
+    - name: Workaround - wait until the 'posthog' pods are ready
+      timeout-minutes: 15
```
tiina303, line 52: nit: maybe increase this to 30, sometimes cluster creation/starting Posthog feels quite slow.

.github/workflows/test-install-digital-ocean.yaml
```diff
@@ -0,0 +1,83 @@
...
+      run: |
+        echo "Waiting for pods to be ready..."
+        while ! kubectl wait --for=condition=Ready pods --timeout=60s --all -n posthog > /dev/null 2>&1
```
tiina303, line 55: I'm not sure, but when we're doing the install we don't seem to start all the pods at the same time (based on looking at `k9s -A`) so there could be a race here, where we just had the ingress first pod ready and continue, but we should wait longer, so maybe the check below should be before or checking for web/events pod specifically too?

tiina303 (approved):
This is great 🙌 

> there's no way atm to run a clean PostHog uninstall. 

Maybe we should in the future  after we fix that add a e2e test that checks that uninstall works properly.

guidoiaquinti:
> This is great 🙌
> 
> > there's no way atm to run a clean PostHog uninstall.
> 
> Maybe we should in the future after we fix that add a e2e test that checks that uninstall works properly.

you mean https://github.com/PostHog/charts-clickhouse/pull/180  😉 ?

.github/workflows/test-install-digital-ocean.yaml
guidoiaquinti: For what I tested this action usually takes < 3 minutes (see [example](https://github.com/PostHog/charts-clickhouse/runs/4239504054?check_suite_focus=true)). I don't think there's need to increase it to 30 (and 15 is probably too much already).

.github/workflows/test-install-digital-ocean.yaml
guidoiaquinti: We can but I wanted to avoid hardcoding pod roles/names into the tests. I'm inclined to keep this till we understand it's a real problem (aka test being flaky for pods not being available)

PR #189, opened 2021-11-16
Improvements to the "Test chart" suite

guidoiaquinti:
## Description
This PR makes a couple of changes to the `Test chart` suite and a macro change in our helm chart K8s version compatibility:

1. the most important is that we weren't actually testing this chart against multiple k8s versions. I've now removed 1.22 from the tests as it's currently failing (the ClickHouse operator is using a deprecated API)

1. the test now creates and uses the `posthog` namespace (like we suggest in our docs)
 
1. the k6 ingestion test is now more configurable as we can now have separate endpoints for the capture and API stack. This will be useful when we'll do ingestion testing also on other platforms (e.g. DigitalOcean) when we'll hit a load balancer and not a service ClusterIP.

## Type of change

## How has this been tested?
CI should be passing.

## Checklist

### DIFF
--- a/.github/workflows/test-helm-chart.yaml
+++ b/.github/workflows/test-helm-chart.yaml
@@ -56,7 +56,7 @@ jobs:
       - name: (Upgrade) Install our helm chart from main branch
         if: matrix.test == 'upgrade'
         run: |
-          helm upgrade --install -f ci/values.yaml --timeout 10m --create-namespace --namespace posthog posthog ./main/charts/posthog
+          helm upgrade --install -f ci/values.yaml --timeout 10m --create-namespace --namespace posthog posthog ./main/charts/posthog --wait --wait-for-jobs
 
       - name: "(Upgrade) Install helm diff"
         if: matrix.test == 'upgrade'
@@ -77,7 +77,7 @@ jobs:
 
       - name: Install our helm chart
         run: |
-          helm upgrade --install -f ci/values.yaml --timeout 10m --create-namespace --namespace posthog posthog ./charts/posthog
+          helm upgrade --install -f ci/values.yaml --timeout 10m --create-namespace --namespace posthog posthog ./charts/posthog --wait --wait-for-jobs
 
       # GitHub Action reference: https://github.com/jupyterhub/action-k8s-await-workloads
       - name: Await all deployments to start
--- a/.github/workflows/test-helm-chart.yaml
+++ b/.github/workflows/test-helm-chart.yaml
@@ -17,12 +17,7 @@ jobs:
         # We run this job multiple times with different parameterization
         # specified below, these parameters have no meaning on their own and
         # gain meaning on how job steps use them.
-        #
-        # k3s-version: https://github.com/rancher/k3s/tags
-        # k3s-channel: https://update.k3s.io/v1-release/channels
         include:
-          - k3s-channel: v1.22
-            test: install
           - k3s-channel: v1.21
             test: install
           - k3s-channel: v1.20
@@ -30,8 +25,8 @@ jobs:
           - k3s-channel: v1.19
             test: install
 
-          # We run two upgrade tests where we first install an already released
-          # Helm chart version and then upgrades to the version we are now
+          # We run an upgrade test where we first install an already released
+          # Helm chart version and then upgrade to the version we are now
           # testing. We test upgrading from the latest main version.
           #
           # It can be very useful to see the "Helm diff" step's output from the
@@ -48,10 +43,9 @@ jobs:
       # ref: https://github.com/jupyterhub/action-k3s-helm/
       - uses: jupyterhub/action-k3s-helm@v1
         with:
-          k3s-channel: v1.19
+          k3s-channel: ${{ matrix.k3s-channel }}
           metrics-enabled: false
           traefik-enabled: false
-          docker-enabled: true
 
       - uses: actions/checkout@v2
         if: matrix.test == 'upgrade'
@@ -62,7 +56,7 @@ jobs:
       - name: (Upgrade) Install our helm chart from main branch
         if: matrix.test == 'upgrade'
         run: |
-          helm upgrade --install posthog main/charts/posthog --timeout 20m -f ci/values.yaml --debug
+          helm upgrade --install -f ci/values.yaml --timeout 10m --create-namespace --namespace posthog posthog ./main/charts/posthog
 
       - name: "(Upgrade) Install helm diff"
         if: matrix.test == 'upgrade'
@@ -72,7 +66,7 @@ jobs:
       - name: "(Upgrade) Helm diff ${{ matrix.upgrade-from }} chart with local chart"
         if: matrix.test == 'upgrade'
         run: |
-          helm diff upgrade --install posthog charts/posthog -f ci/values.yaml --debug
+          helm diff upgrade --install -f ci/values.yaml --namespace posthog posthog ./charts/posthog
 
       - name: "(Upgrade) Await chart install"
         if: matrix.test == 'upgrade'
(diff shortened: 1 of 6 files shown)


.github/workflows/test-helm-chart.yaml
```diff
@@ -92,16 +92,16 @@ jobs:
           timeout: 900
           max-restarts: 15
 
-      - name: Setup for ingestion test
-        run: |
-          export WEB_POD=$(kubectl get pods -l role=web -o jsonpath="{.items[].metadata.name}")
-          kubectl exec $WEB_POD -- python manage.py setup_dev --no-data
+      - name: Setup PostHog for the ingestion test
+        run: ci/setup_ingestion_test.sh
```
guidoiaquinti, line 96: I've moved this to an external script so other e2e test can use it as well

.github/workflows/test-helm-chart.yaml
```diff
@@ -48,10 +43,9 @@ jobs:
       # ref: https://github.com/jupyterhub/action-k3s-helm/
       - uses: jupyterhub/action-k3s-helm@v1
         with:
-          k3s-channel: v1.19
```
guidoiaquinti, line 51: this is probably the most important fix of this PR

.github/workflows/test-helm-chart.yaml
```diff
@@ -48,10 +43,9 @@ jobs:
...
+          k3s-channel: ${{ matrix.k3s-channel }}
           metrics-enabled: false
           traefik-enabled: false
-          docker-enabled: true
```
guidoiaquinti, line 54: The action default is `false` and the action seems to run faster without it

ci/values.yaml
```diff
@@ -1,8 +1,5 @@
 cloud: k3s
 
-image:
```
guidoiaquinti, line 3: Not sure why we want to test against `latest` but this makes the test non deterministic. I'm removing this so we can use the chart default (latest stable release)

ci/values.yaml
```diff
@@ -12,16 +9,8 @@ ingress:
     enabled: false
 
 clickhouseOperator:
-  enabled: true
-  namespace: default
   storage: 2Gi
 
-plugins:
```
guidoiaquinti, line 19: Those are already the defaults

.github/workflows/test-helm-chart.yaml
macobo: :facepalm:

.github/workflows/test-helm-chart.yaml
```diff
@@ -92,16 +86,16 @@ jobs:
...
 
-          echo "POSTHOG_WEB_HOSTNAME=$(kubectl get svc posthog-web -o jsonpath="{.spec.clusterIP}")" >> $GITHUB_ENV
-          echo "POSTHOG_EVENTS_HOSTNAME=$(kubectl get svc posthog-events -o jsonpath="{.spec.clusterIP}")" >> $GITHUB_ENV
+      - name: Fetch PostHog endpoints to use for the ingestion test
+        run: |
+          POSTHOG_API_ADDRESS=$(kubectl get svc -n posthog posthog-web -o jsonpath="{.spec.clusterIP}")
+          POSTHOG_EVENTS_ADDRESS=$(kubectl get svc -n posthog posthog-events -o jsonpath="{.spec.clusterIP}")
 
-          # Sleep until web service is ready to respond
-          sleep 20
```
macobo, line 104: Will the endpoint be ready to respond immediately? I wouldn't expect helm upgrade to block until then.

ci/k6-ingestion-test.js
```diff
@@ -1,6 +1,18 @@
 import http from 'k6/http'
-import {check, group, sleep} from 'k6'
+import {check, sleep, fail} from 'k6'
 import { Gauge } from 'k6/metrics'
+import { URL } from 'https://jslib.k6.io/url/1.0.0/index.js';
+
+const POSTHOG_ENDPOINT = __ENV.POSTHOG_ENDPOINT
+const POSTHOG_API_ENDPOINT = __ENV.POSTHOG_API_ENDPOINT || POSTHOG_ENDPOINT
+const POSTHOG_EVENT_ENDPOINT = __ENV.POSTHOG_EVENT_ENDPOINT || POSTHOG_ENDPOINT
```
macobo, line 8: I don't think the extra configuration here is warranted (YAGNI), or am I missing something?

.github/workflows/test-helm-chart.yaml
macobo: Okay, so the guarantee is that after `helm install` the site is immediately accessible? That's surprising for me and wouldn't work with e.g. DNS.

ci/k6-ingestion-test.js
```diff
@@ -1,6 +1,17 @@
...
 import { Gauge } from 'k6/metrics'
+import { URL } from 'https://jslib.k6.io/url/1.0.0/index.js';
+
+const POSTHOG_API_ENDPOINT = __ENV.POSTHOG_API_ENDPOINT
+const POSTHOG_EVENT_ENDPOINT = __ENV.POSTHOG_EVENT_ENDPOINT
+
+const captureURL = new URL(`${POSTHOG_EVENT_ENDPOINT}/e/`);
+const apiURL = new URL(`${POSTHOG_API_ENDPOINT}/api/insight/trend/?events=[{"id":"k6s_custom_event","type":"events"}]&refresh=true`);
+
+if (POSTHOG_EVENT_ENDPOINT == null && POSTHOG_API_ENDPOINT == null) {
```
macobo, line 12: Suggestion:

```
if (!POSTHOG_EVENT_ENDPOINT || !POSTHOG_API_ENDPOINT) {
```

If either is missing you want to fail, as with empty strings.

ci/k6-ingestion-test.js
```diff
@@ -1,6 +1,17 @@
 import http from 'k6/http'
-import {check, group, sleep} from 'k6'
+import {check, sleep, fail} from 'k6'
 import { Gauge } from 'k6/metrics'
+import { URL } from 'https://jslib.k6.io/url/1.0.0/index.js';
+
+const POSTHOG_API_ENDPOINT = __ENV.POSTHOG_API_ENDPOINT
+const POSTHOG_EVENT_ENDPOINT = __ENV.POSTHOG_EVENT_ENDPOINT
+
+const captureURL = new URL(`${POSTHOG_EVENT_ENDPOINT}/e/`);
```
macobo, line 9: Note: There's now very inconsistent semi-colon usage through the file.

Worth adding eslint to the project to autoformat this or just removing?

ci/k6-ingestion-test.js
```diff
@@ -1,6 +1,17 @@
 import http from 'k6/http'
-import {check, group, sleep} from 'k6'
+import {check, sleep, fail} from 'k6'
 import { Gauge } from 'k6/metrics'
+import { URL } from 'https://jslib.k6.io/url/1.0.0/index.js';
```
macobo, line 4: This import is weird. Context?

.github/workflows/test-helm-chart.yaml
guidoiaquinti: The guarantee is that after helm install the resources are installed, this doesn't imply the site is immediately accessible tho (e.g. if the app takes 20 second to start). In this case anyway we only need to fetch the endpoint from the resources so we should be good.

ci/k6-ingestion-test.js
guidoiaquinti: I can move it back to string if you prefer. I was following the [docs](https://k6.io/docs/examples/urls-with-query-parameters/)

macobo: Nit: Last remaining semicolon :D

guidoiaquinti: 🤦

PR #196, opened 2021-11-18
e2e - DigitalOcean (install)

guidoiaquinti:
## Description
1. This PR brings an e2e test to deploy PostHog via Helm on DigitalOcean. 

    It includes certificate signing using Let's Encrypt (see [here](https://crt.sh/?q=posthog.cc) the certificate transparency log entries). The run times are not great as it takes between 10 and 30 min only between DNS record creation and TLS record deployment but once I understand why DNS propagation is so slow we might be able to speed up the process (e.g. by using a different validation method).

2. As part of this PR I'm enhancing but also disabling the `e2e - DigitalOcean 1-click install` as it's currently flaky due to some issues (most probably) related to the DO marketplace. I'm actively investigating the issue but I would appreciate if we can merge this so I can move with e2e tests for other platforms.

## Type of change

## How has this been tested?
CI should be passing.

## Checklist


.github/workflows/test-digital-ocean-1-click.yaml
```diff
@@ -1,10 +1,18 @@
-name: e2e - DigitalOcean 1-click (install)
-
-on: push
+name: e2e - DigitalOcean (1-click)
+#
+# Unfortunately we can test this scenario only after merging to 'main' as we don't have
```
guidoiaquinti, line 3: I've realised about this limitation while working on the non 1-click install. I'll try to come up with a workaround as this situation is not ideal.

.github/workflows/test-digital-ocean-1-click.yaml
```diff
@@ -59,24 +60,11 @@ jobs:
...
-        namespace: "posthog"
-        timeout: 300
-        max-restarts: 10
-
     # TODO
     # - setup ingestion test
     # - run ingestion test
     # - run k8s spec test
 
-    - name: Emit k8s namespace report
```
guidoiaquinti, line 74: I didn't find this reporting useful. I think it can be superseded by  k8s spec test

.github/workflows/test-install-digital-ocean.yaml
```diff
@@ -1,83 +0,0 @@
-name: e2e - DigitalOcean 1-click (install)
```
guidoiaquinti, line 1: File has been renamed

.github/workflows/test-digital-ocean-1-click-install.yaml
```diff
@@ -0,0 +1,113 @@
...
+#
+# Unfortunately we can test this scenario only after merging to 'main' as we don't have
+# control over the branch used by the DigitalOcean's marketplace 1-click install.
+#
+# ref: https://github.com/digitalocean/marketplace-kubernetes/blob/master/stacks/posthog/deploy.sh
+#
+on:
+  push:
+    # branches:
+    #   - main
```
guidoiaquinti, line 20: I've left this commented to test this action in this PR. I'm going to uncomment it once I have your ✅  and I see the test passing.

.github/workflows/test-digital-ocean-1-click-install.yaml
```diff
@@ -0,0 +1,113 @@
...
+      uses: jupyterhub/action-k8s-await-workloads@main
+      with:
+        namespace: "posthog"
+        timeout: 300
+        max-restarts: 10
+
+    - name: Setup PostHog for the ingestion test
+      run: ./ci/setup_ingestion_test.sh
+
+    - name: Set PostHog endpoints to use for the ingestion test
```
guidoiaquinti, line 97: I've added the ingestion test to this scenario

.github/workflows/test-digital-ocean-1-click.yaml
tiina303: briefly discussed in our 1:1 that we could use an env variable that's read in `deploy.sh` & define it after we create the DO cluster, then run the deployment script

.github/workflows/test-digital-ocean-install.yaml
```diff
@@ -0,0 +1,141 @@
...
+          posthog.cc \
+          --record-type A \
+          --record-ttl 60 \
+          --record-name "${{ steps.vars.outputs.dns_record }}" \
+          --record-data "$load_balancer_external_ip"
+
+    - name: Wait for the Let's Encrypt certificate to be issued and deployed
+      run: |
+        echo "Wait for the Let's Encrypt certificate to be issued and deployed..."
+        while ! kubectl wait --for=condition=Ready --timeout=60s certificaterequest --all -n posthog > /dev/null 2>&1
```
tiina303, line 113: we should likely add a timeout here too (similar to above)?

tiina303 (approved):
looks good  👍

.github/workflows/test-digital-ocean-install.yaml
guidoiaquinti: Fair, my take is that we could leave this to fail via the overall action timeout

Issue #174, opened 2021-11-02
Don't install unnecessary `StorageClass`es


guidoiaquinti:
## Is your feature request related to a problem?

As part of the `clickhouse_instance.yaml` template definition, if we are installing the chart on GCP or AWS, we add (by default) an unnecessary `StorageClass` definition.

```
kubectl get storageclass

NAME            PROVISIONER             RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
gp2 (default)   kubernetes.io/aws-ebs   Delete          WaitForFirstConsumer   false                  15h
gp2-resizable   kubernetes.io/aws-ebs   Retain          Immediate              true                   15h
```

This `StorageClass` is then only referenced by `ClickHouseInstallation` and not by other stateful components. I think we should remove this unnecessary complication especially as it created confusions in the past.  A `StorageClass` is also a k8s global resource, we might hit naming collisions, etc… 

Unfortunately this will be a breaking change and it will require a safe migration strategy like:

1. make sure `helm` runs cleanly
1. make sure the default `StorageClass` has `VolumeExpansion` set to true
1. stop the ClickHouse pod (make sure we don't destroy the underlying PVC/PV)
1. use something like `pv-migrate` (see https://github.com/utkuozdemir/pv-migrate) to migrate the data from the source PV to the destination PV.
1. delete the `ClickHouseInstallation` definition
1. manually patch the `PersistentVolume` definition to use the new volumes
1. run `helm` upgrade with the new config to create a new `ClickHouseInstallation` definition

## Describe the solution you’d like
By default, PostHog shouldn't require any custom `StorageClass`.

## Describe alternatives you’ve considered
Do nothing.

tiina303:
If we want to do this, it would be great if we could do this before getting new installs from product-hunt or (perhaps we can temporarily make new installs not have storage class and existing installs keep the storage class for now)?

Legion2:
Also related to https://github.com/PostHog/charts-clickhouse/issues/215

guidoiaquinti:
Closed via #224

PR #197, opened 2021-11-18
Clickhouse encrypted drive

tiina303:
Will be used for a custom upgrade. Will not ship.

### DIFF
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -21,9 +21,13 @@ apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
   name: gp2-resizable
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"   
 provisioner: kubernetes.io/aws-ebs
 parameters:
   type: gp2
+  encrypted: 'true'
+  fsType: ext4
 reclaimPolicy: Retain
 #volumeBindingMode: Immediate
 allowVolumeExpansion: true


charts/posthog/templates/clickhouse_instance.yaml
```diff
@@ -21,9 +21,13 @@ apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
   name: gp2-resizable
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
```
Legion2, line 25: there can/should only be one default class per cluster. A normal user of this chart would not expect that the chart adds a default storage class, because setting up storage classes is what a cluster administrator does and not something an application administator would do.

Also creating non namespaced resources which are hidden in the helm chart are not best practice, because this can result in conflicts with existing resources in the cluster.

Legion2 (changes requested):

charts/posthog/templates/clickhouse_instance.yaml
tiina303: Thanks. Note that this PR wasn't intended to be landed just to be clear. It was only used for a custom upgrade.

guidoiaquinti: +1 to what @Legion2 wrote. See also https://github.com/PostHog/charts-clickhouse/issues/174 for more info

guidoiaquinti: Can we close this without merging @tiina303 ?

guidoiaquinti:
@tiina303 I'm closing this, feel free to re-open if I've missed anything. Thank you!

PR #198, opened 2021-11-18
enable TLS connection from pgbouncer to hosted PG

fuziontech:
## Description
This is required if you use hosted PostgreSQL like RDS

### DIFF
--- a/charts/posthog/templates/_helpers.tpl
+++ b/charts/posthog/templates/_helpers.tpl
@@ -133,7 +133,7 @@ Set postgres password b64
 Set postgres URL
 */}}
 {{- define "posthog.postgresql.url" -}}
-    postgres://{{- .Values.postgresql.postgresqlUsername -}}:{{- template "posthog.postgresql.password" . -}}@{{- template "posthog.postgresql.host" .  -}}:{{- template "posthog.postgresql.port" . -}}/{{- .Values.postgresql.postgresqlDatabase }}
+    postgres://{{- .Values.postgresql.postgresqlUsername -}}:{{- template "posthog.postgresql.password" . -}}@{{- template "posthog.postgresql.host" .  -}}:{{- template "posthog.postgresql.port" . -}}/{{- .Values.postgresql.postgresqlDatabase -}}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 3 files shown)


charts/posthog/values.yaml
```diff
@@ -344,6 +344,8 @@ ingress:
 postgresql:
   # -- Install postgres server on kubernetes (see below)
   enabled: true
+  # -- sslmode for connection - defaults to prefer
```
macobo, line 347: Worth adding to the comment, what are the valid values?

sjmadsen:
This PR has been open for over two months and is a blocker for externally hosted Postgres. Is there a plan for when it will be merged?

charts/posthog/templates/_helpers.tpl
```diff
@@ -133,7 +133,7 @@ Set postgres password b64
 Set postgres URL
 */}}
 {{- define "posthog.postgresql.url" -}}
-    postgres://{{- .Values.postgresql.postgresqlUsername -}}:{{- template "posthog.postgresql.password" . -}}@{{- template "posthog.postgresql.host" .  -}}:{{- template "posthog.postgresql.port" . -}}/{{- .Values.postgresql.postgresqlDatabase }}
+    postgres://{{- .Values.postgresql.postgresqlUsername -}}:{{- template "posthog.postgresql.password" . -}}@{{- template "posthog.postgresql.host" .  -}}:{{- template "posthog.postgresql.port" . -}}/{{- .Values.postgresql.postgresqlDatabase -}}
```
tiina303, line 136: don't we need the new line in the end here?

charts/posthog/values.yaml
tiina303: pretty easy to find on Google and that could change in theory if anything link to https://www.pgbouncer.org/config.html ?

tiina303 (approved):
fix templates file before merging

guidoiaquinti:
> This PR has been open for over two months and is a blocker for externally hosted Postgres. Is there a plan for when it will be merged?

@sjmadsen you can already use an externally hosted PostgreSQL using TLS by simply passing custom env variables to the pgbouncer deployment via `pgbouncer.env`. Example:

```yaml
pgbouncer:
  env:
  - name: SERVER_TLS_SSLMODE
    value: "your_value"
```

This PR only makes explicit the TLS configuration. @hazzadous I honestly don't see the value of adding this explicit variable but maybe I am missing something. Are you 👍 if I close without merging?

hazzadous:
> Are you 👍 if I close without merging?

I'm ok with that. We can make sure setting the env is documented.

Issue #194, opened 2021-11-17
Helm fails to render the chart when Redis is disabled


aleksanderllada:
## Bug description
I tried to install the Helm chart using an external Redis cluster and it failed with the following error message:

```bash
helm install -f posthog.yaml --timeout 20m -n applications posthog posthog/posthog --dry-run --debug --version=5.1.0
install.go:159: [debug] Original chart version: "5.1.0"
install.go:176: [debug] CHART PATH: /Users/aleksanderarruda/Library/Caches/helm/repository/posthog-5.1.0.tgz

install.go:184: [debug] WARNING: This chart or one of its subcharts contains CRDs. Rendering may fail or contain inaccuracies.
Error: YAML parse error on posthog/templates/plugins-deployment.yaml: error converting YAML to JSON: yaml: line 61: did not find expected key
helm.go:84: [debug] error converting YAML to JSON: yaml: line 61: did not find expected key
```

## How to reproduce
1. Use the following values.yaml file:
```bash
cloud: "aws"

redis:
  enabled: false

  host: "redis.com"
  password: "doesntmatter"
  port: "6379"
```
2. Install the Helm chart (`--dry-run` should be enough):
```bash
helm install -f posthog.yaml --timeout 20m -n applications posthog-clickhouse posthog/posthog --dry-run --debug --version=5.1.0
```

## Environment

- Deployment platform: AWS
- Chart version/commit: 5.1.0
- Posthog version: 1.30.0

tiina303:
The error happens because we try to set the env
```
 - name: REDIS_URL
          value: "redis://:$(POSTHOG_REDIS_PASSWORD)@"redis.com":6379"
```
which isn't valid yaml.

guidoiaquinti:
Closing as #248 is now merged. @aleksanderllada feel free to re-open or ping me if something doesn't work as expected. Thank you!

PR #199, opened 2021-11-19
Fix redis url bug

tiina303:
## Description

Closes https://github.com/PostHog/charts-clickhouse/issues/194

## Type of change

## How has this been tested?
```
helm template -f customer.values.yaml posthog ~/Posthog/charts-clickhouse/charts/posthog | grep "REDIS_URL" -A2
        - name: REDIS_URL
          value: "redis://:$(POSTHOG_REDIS_PASSWORD)@$(POSTHOG_REDIS_HOST):$(POSTHOG_REDIS_PORT)"
        - name: CLICKHOUSE_DATABASE
```

## Checklist

### DIFF
--- a/charts/posthog/templates/plugins-deployment.yaml
+++ b/charts/posthog/templates/plugins-deployment.yaml
@@ -89,16 +89,16 @@ spec:
           value: {{ template "posthog.pgbouncer.url" . }}
         - name: USING_PGBOUNCER
           value: 'true'
-          {{- if or (.Values.redis.enabled) (.Values.redis.password) }}
-        - name: POSTHOG_REDIS_PASSWORD
-          value: {{ .Values.redis.password | quote }}
-        - name: REDIS_URL
-          value: "redis://:$(POSTHOG_REDIS_PASSWORD)@{{- template "posthog.redis.host" . -}}:{{-  template "posthog.redis.port" . -}}"
-        {{- end }}
         - name: POSTHOG_REDIS_HOST
           value: {{ template "posthog.redis.host" . }}
         - name: POSTHOG_REDIS_PORT
           value: {{ include "posthog.redis.port" . | quote }}
+          {{- if or (.Values.redis.enabled) (.Values.redis.password) }}
+        - name: POSTHOG_REDIS_PASSWORD
+          value: {{ .Values.redis.password | quote }}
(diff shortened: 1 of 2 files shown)


guidoiaquinti:
Can we add a test to make sure we don't introduce a regression in the future?

charts/posthog/templates/plugins-deployment.yaml
```diff
@@ -89,16 +89,16 @@ spec:
           value: {{ template "posthog.pgbouncer.url" . }}
         - name: USING_PGBOUNCER
           value: 'true'
+        - name: POSTHOG_REDIS_HOST
+          value: {{ template "posthog.redis.host" . }}
+        - name: POSTHOG_REDIS_PORT
+          value: {{ include "posthog.redis.port" . | quote }}
```
guidoiaquinti, line 95: Why do we need the `| quote` here  while for the host we don't?

charts/posthog/tests/redis.yaml
```diff
@@ -0,0 +1,63 @@
+suite: Redis
```
guidoiaquinti, line 1: If we call the suite `Redis` I think we should include here all the possible rendered values used by the chart. 
If we call this `Plugins deployment definition` I think it could work (despite we don't test a lot of other stuff related to it)

guidoiaquinti: Iterating on this comment: why do we have `REDIS_URL` only for the plugins deployment and not for events/migrate/beat/web and workers?

guidoiaquinti:
Few comments but overall LGTM

charts/posthog/templates/plugins-deployment.yaml
tiina303: I just moved it, no idea what should it be?

charts/posthog/tests/redis.yaml
tiina303: I'm not sure how we use Redis exactly, but I suspect we should have it set for web & events & beat too based on https://posthog.com/docs/self-host/architecture.

guidoiaquinti:
I'm closing the PR as https://github.com/PostHog/charts-clickhouse/pull/248 supersedes it.

PR #200, opened 2021-11-23
e2e - Google Cloud Platform (install)

guidoiaquinti:
## Description
Similar to https://github.com/PostHog/charts-clickhouse/pull/196 this PR brings an e2e test to deploy PostHog via Helm on Google Cloud Platform (as requested in #183).

It includes certificate signing using Google-managed SSL certificates (see [here](https://crt.sh/?q=posthog.cc) for the certificate transparency log entries).

Due to a bug in the [GCE ingress](https://stackoverflow.com/questions/44584270/how-to-get-a-custom-healthcheck-path-in-a-gce-l7-balancer-serving-a-kubernetes-i) that I still need to work around, I'm temporary disabling the k6 test as the test will fail otherwise.

I'm inclined to merge this anyway as it is and then revisit the k6 test once the e2e tests for the other platforms are also shipped.

## Type of change

## How has this been tested?
CI should be passing.

## Checklist

### DIFF
--- a/.github/workflows/test-digitalocean-install.yaml
+++ b/.github/workflows/test-digitalocean-install.yaml
@@ -7,7 +7,7 @@
 #
 name: e2e - DigitalOcean (install)
 
-on: push
+on: [push, workflow_dispatch]
 
 jobs:
   do-install:
@@ -18,7 +18,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: Install doctl
+    - name: Install doctl to manage 'posthog.cc' DNS
       uses: digitalocean/action-doctl@v2
       with:
         token: ${{ secrets.DIGITALOCEAN_ACCESS_TOKEN }}
--- a/.github/workflows/test-digitalocean-install.yaml
+++ b/.github/workflows/test-digitalocean-install.yaml
@@ -81,8 +81,8 @@ jobs:
         load_balancer_external_ip=""
         while [ -z "$load_balancer_external_ip" ];
         do
+          echo "  sleeping 10 seconds" && sleep 10
           load_balancer_external_ip=$(kubectl get ingress -n posthog posthog -o jsonpath="{.status.loadBalancer.ingress[0].ip}")
-          [ -z "$load_balancer_external_ip" ] && echo "  sleeping 10 seconds" && sleep 10
         done
         echo "The DigitalOcean Load Balancer is now ready!"
 
@@ -108,6 +108,7 @@ jobs:
           --record-data "$load_balancer_external_ip"
 
     - name: Wait for the Let's Encrypt certificate to be issued and deployed
+      id: tls_certificate_creation
       run: |
         echo "Wait for the Let's Encrypt certificate to be issued and deployed..."
         while ! kubectl wait --for=condition=Ready --timeout=60s certificaterequest --all -n posthog > /dev/null 2>&1
@@ -129,13 +130,19 @@ jobs:
       with:
         filename: ci/k6-ingestion-test.js
 
-    - name: Delete the k8s cluster and all associated resources (LB, volumes, ...)
+    - name: Delete the k8s cluster and all the associated resources
       if: ${{ always() && steps.k8s_cluster_creation.outcome == 'success' }}
       run: |
-        doctl k8s cluster delete --dangerous --force ${{ steps.vars.outputs.k8s_cluster_name }}
+        doctl k8s cluster delete \
+          --dangerous \
+          --force \
+          ${{ steps.vars.outputs.k8s_cluster_name }}
 
     - name: Delete the DNS record
       if: ${{ always() && steps.dns_creation.outcome == 'success' }}
       run: |
         DNS_RECORD_ID=$(doctl compute domain records list posthog.cc --no-header --format ID,Name | grep ${{ steps.vars.outputs.dns_record }} | awk '{print $1}')
-        doctl compute domain records delete --force posthog.cc "$DNS_RECORD_ID"
+        doctl compute domain records delete \
+          posthog.cc \
+          --force \
+          "$DNS_RECORD_ID"
--- a/.github/workflows/test-google-cloud-platform-install.yaml
+++ b/.github/workflows/test-google-cloud-platform-install.yaml
@@ -155,10 +155,6 @@ jobs:
     #
     # see: https://stackoverflow.com/questions/44584270/
     #
-    - name: Run ingestion test using k6
-      run: |
-        sleep 30
-
     # - name: Run ingestion test using k6
     #   uses: k6io/action@v0.2.0
     #   with:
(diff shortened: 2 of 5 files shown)


ci/k6-ingestion-test.js
```diff
@@ -17,8 +17,8 @@ let eventsIngested = new Gauge('events_ingested')
 
 export let options = {
   thresholds: {
-    http_req_failed: ['rate<0.01'],   // http errors should be less than 1%
-    http_req_duration: ['p(95)<500'], // 95% of requests should be below 500ms
+    http_req_failed: ['rate < 0.01'],   // http errors should be less than 1%
+    http_req_duration: ['p(90) < 1000'], // 90% of requests should be below 1s
```
guidoiaquinti, line 21: Not sure if it's the new PostHog build but latency seems to have increased

macobo: If this continues, consider making this into an env var to make it elaborate why it's failing there?

.github/workflows/test-google-cloud-platform-install.yaml
```diff
@@ -0,0 +1,200 @@
...
+    - name: Setup PostHog for the ingestion test
+      run: ./ci/setup_ingestion_test.sh
+
+    - name: Set PostHog endpoints to use for the ingestion test
+      run: |
+        echo "POSTHOG_API_ENDPOINT=https://${{ steps.vars.outputs.fqdn_record }}" | tee -a "$GITHUB_ENV"
+        echo "POSTHOG_EVENT_ENDPOINT=https://${{ steps.vars.outputs.fqdn_record }}" | tee -a "$GITHUB_ENV"
+
+    #
+    # TODO: the GCE Ingress is not picking up health check from readiness probe definition and it's using '/'
```
macobo, line 153: Could we add a sleep timer temporarily? Feels better than not doing anything.

guidoiaquinti:
> How much will this $$$? :)

The test runs in ~30 min: 6 nodes `e2-medium` (2 nodes * 3 AZs) * $0.03351 on demand cost / h =  ~0.10 USD

.github/workflows/test-google-cloud-platform-install.yaml
```diff
@@ -0,0 +1,204 @@
+#
+# This is an e2e test to deploy PostHog on Google Cloud Platform using Helm.
+#
+# TODO:
+# - run k8s spec test
+# - run action only when necessary
+#
+name: e2e - Google Cloud Platform (install)
+
+on: push
```
tiina303, line 10: can we add `workflow-dispatch` to all of these too, just so we can trigger them from actions if we want to

.github/workflows/test-google-cloud-platform-install.yaml
```diff
@@ -0,0 +1,204 @@
...
+      uses: actions/checkout@v2
+
+    - name: Set up GCP Cloud SDK
+      uses: google-github-actions/setup-gcloud@master
+      with:
+        project_id: ${{ secrets.GCP_PROJECT_ID }}
+        service_account_key: ${{ secrets.GCP_SA_KEY }}
+        export_default_credentials: true
+
+    - name: Install doctl
```
tiina303, line 28: consider:
```suggestion
    - name: Install doctl for DNS
```

.github/workflows/test-google-cloud-platform-install.yaml
```diff
@@ -0,0 +1,204 @@
...
+        echo "::set-output name=dns_record::${TEST_NAME}"
+        echo "::set-output name=fqdn_record::${TEST_NAME}.posthog.cc"
+
+    - name: Deploy a new k8s cluster
+      id: k8s_cluster_creation
+      run: |
+        gcloud container clusters create \
+          ${{ steps.vars.outputs.k8s_cluster_name }} \
+          --project ${{ secrets.GCP_PROJECT_ID }} \
+          --region europe-west4 \
```
tiina303, line 47: lets use `us-central-1` as that's cheaper and we don't care where we deploy. See https://cloud.google.com/products/calculator#id=ee71f5f9-4a0a-4caa-a122-e9389cdb8f3e

.github/workflows/test-google-cloud-platform-install.yaml
```diff
@@ -0,0 +1,204 @@
...
+        echo "Waiting for the GCP Load Balancer to be ready..."
+        load_balancer_external_ip=""
+        while [ -z "$load_balancer_external_ip" ];
+        do
+          echo "  sleeping 10 seconds" && sleep 10
+          load_balancer_external_ip=$(kubectl get ingress -n posthog posthog -o jsonpath="{.status.loadBalancer.ingress[0].ip}")
+        done
+        echo "The GCP Load Balancer is now ready!"
+
+    - name: Wait until all the resources are fully deployed in k8s
```
tiina303, line 125: Could we use sth like https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/ to reduce the code duplication

tiina303 (changes requested):
$$$ note that there's charges for lb, ip, dns, storage too, but all that's still probably close to 0.1-0.2$

.github/workflows/test-google-cloud-platform-install.yaml
guidoiaquinti: Yes and that's part of the plan. I'm going to do that once all the steps are fully fleshes on all platforms

.github/workflows/test-google-cloud-platform-install.yaml
guidoiaquinti: Yes, that's by design. We can't enable it till we add a workaround for the ingress healthcheck bug

ci/k6-ingestion-test.js
guidoiaquinti: Looking at the action perf history, this issue seems to happen only on DO where we use very basic droplets with shared resources.

.github/workflows/test-digitalocean-install.yaml
```diff
@@ -7,7 +7,7 @@
 #
 name: e2e - DigitalOcean (install)
 
-on: push
+on: [push, workflow_dispatch]
```
macobo, line 10: Note that most other workflows we have (I'd use posthog/ repo as a canonical example) uses pull_request trigger.

`workflow_dispatch` feels like YAGNI but w/e :)

macobo (approved):
Code-wise this looks good, but no idea strategy-wise.

.github/workflows/test-digitalocean-install.yaml
guidoiaquinti: I've added `workflow_dispatch` as requested by @tiina303 [here](https://github.com/PostHog/charts-clickhouse/pull/200#discussion_r755463606). 

Regarding `pull_request` VS `push` we can revisit this if we think CI checks will run too often.

guidoiaquinti:
> Code-wise this looks good, but no idea strategy-wise.

Thank you @macobo ! I'll be happy to get your strategy-wise view as well

PR #202, opened 2021-11-25
Unit test / Lint test - run only when necessary

guidoiaquinti:
## Description
1. split the lint action into 2 so each test can run independently and only when necessary
2. made some changes to the unit test action so it can run only when necessary

## Type of change

## How has this been tested?
CI is passing and it's triggered when expected

## Checklist


tiina303 (approved):
🚢

Issue #132, opened 2021-09-27
IP address captured is worker node IP instead of user IP


rameshlohala:
## Bug description

The IP address captured in the events is Kubernetes worker node instead of the user's IP address. 

I had also tried enabling proxy protocol to ELB following this documentation https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html

Posthog is deployed on AWS EKS.

![IP](https://user-images.githubusercontent.com/9413979/134872318-328bb6d4-2f13-4382-9610-24170ce6fd9e.png)

## Expected behavior
1. IP address captured in the events should be user's public IP

## Environment

- Deployment platform: aws
- Chart version/commit: 3.12.0
- Posthog version: 1.28.0

guidoiaquinti:
👋 Hi @rameshlohala, thank you for your report. 

I was able to reproduce your issue and it seems to be due to how the AWS ELB (classic) works. In order to fix the issue, can you try to:

1. enable the `PROXY` protocol for the ELB (see official [AWS documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html#enable-proxy-protocol-cli))
2. add this snippet to your `values.yaml` and run a `helm upgrade`:
    ```
    # For AWS ELB in L4 (TCP) mode, we need to enable some additional config
    # in the ingress controller in order to get the proper IP address forwarded
    # to our app. Otherwise we'll get the load balancer nodes addresses instead.
    #
    # See: https://kubernetes.github.io/ingress-nginx/user-guide/miscellaneous/#source-ip-address
    ingress-nginx:
      controller:
        config:
          use-proxy-protocol: true
    ```

?

It should fix your issue.

rameshlohala:
@guidoiaquinti I have followed the process that you mentioned above and it is fixed now. Thanks

guidoiaquinti:
See also https://github.com/PostHog/posthog.com/pull/2554

PR #208, opened 2021-11-30
Fix - GeoIP does not work on self-hosted with ingress.nginx

guidoiaquinti:
## Description

As reported in https://github.com/PostHog/posthog/issues/6334 and https://github.com/PostHog/charts-clickhouse/issues/132 the GeoIP plugin does not work on self-hosted deployments when `ingress.nginx` is enabled.

This PR fixes the issue by:
1. adding metadata annotations to the `ingress-nginx` service definition in order to enable LB proxy mode for AWS and DO deployments (we will validate Azure and GCP in another PR as their e2e tests are currently blocked)

3. enhancing the current integration test suite we have in CI. The major change is that we added a new test to verify if the source IP address of a random ingested event is part of a public range. This will hopefully catch any regression in the future.

## Type of change

## How has this been tested?
CI should be passing

## Checklist


ci/values/amazon_web_services.yaml
```diff
@@ -5,3 +5,17 @@ ingress:
     enabled: true
 cert-manager:
   enabled: true
+
+#
+# For AWS ELB in L4 (TCP) mode, we need to enable some additional config
```
guidoiaquinti, line 10: Those will eventually be moved as default when `ingress.nginx.enabled` is set to `true`

ci/values/digital_ocean.yaml
```diff
@@ -5,3 +5,17 @@ ingress:
     enabled: true
 cert-manager:
   enabled: true
+
```
guidoiaquinti, line 8: Ditto

charts/posthog/tests/ingress.yaml
```diff
@@ -32,3 +32,115 @@ tests:
           # Unfortunately we can't match the whole manifest as there are few
           # metadata values linked to the chart version
           path: spec
+
```
guidoiaquinti, line 35: This is not strictly part of this PR but it's just some additional unit tests for the ingress definition

tiina303 (approved):

tiina303:
we should change https://github.com/PostHog/charts-clickhouse/blob/main/do_1_click_values.yaml too?

guidoiaquinti:
Actually I think we can't make this working without making PostHog install via Helm a two step process. The problem and potential solution are well explained [here](https://github.com/compumike/hairpin-proxy#the-problem) but the tl;dr; is that we can't make the ACME challenge pass when using PROXY protocol due to a limitation in `cert-manager`.

ci/values/amazon_web_services.yaml
guidoiaquinti: I don't think there's an easy way to override those at runtime.

guidoiaquinti:
This is now working 🎉 

I'm waiting DO support to understand if there's a way to get the LB IP address while using the `do-loadbalancer-hostname` annotation without using `doctl` or looking at the DO UI as without it we'll break the handy command we use in our docs:
```
export INGRESS_IP=$(kubectl get --namespace posthog ingress posthog -o jsonpath="{.status.loadBalancer.ingress[0].ip}") && \
echo "\n-----\n" && \
echo "Your PostHog installation is available at: http://$INGRESS_IP" && \
echo "\n-----\n"
```

In the meantime, feedback to the PR content are welcome!

.github/workflows/test-digitalocean-install.yaml
```diff
@@ -13,7 +13,7 @@ on:
     paths:
       - ci/**
       - charts/**
-      - .github/workflows/test-digital-ocean-install.yaml
+      - .github/workflows/test-digitalocean-install.yaml
```
fuziontech, line 16: 👍

fuziontech (approved):
This is fantastic - I'll be using a good bit of this for the automation interface as well 🚢

guidoiaquinti:
I'll update the docs but it's unfortunately not possible to get the IP in DO

> Hello, 
> 
> Thank you for your patience. 
> 
> Our Kubernetes Experts provided an update on your question about getting the Load Balancer IP. Using the service.beta.kubernetes.io/do-loadbalancer-hostname the annotation specifies the hostname used for the Service status.Hostname instead of assigning status.IP
> 
> We do not have another way to get the Load Balancers IP address you would need to either use the cloud panel or doctl to get it. 
>
> If you have any further questions please let us know.
>
> Kind Regards, 
>
> Tyler J 
>
> DigitalOcean Customer Success
>

guidoiaquinti:
Merging also if e2e GCP is failing. The issue seems to be unrelated.

Issue #210, opened 2021-12-01
Clickhouse nodeSelector can not be set to a custom value when using the clickhouse operator


Legion2:
## Bug description
Clickhouse nodeSelector can not be set to a custom value if using the clickhouse operator.
The helm chart only allows to set a boolean value for `clickhouseOperator.useNodeSelector`

## Expected behavior

The `nodeSelector` property can be set to a custom list of node selectors

## Actual behavior

Only `clickhouse: true` can be set as node selector.

## How to reproduce

1. try to install the posthog chart with custom node selector for the clickhouse instance

## Environment

- [ ] Deployment platform (AWS/Azure/DigitalOcean/GCP/...): not relevant
- [ ] k8s version (server version from `kubectl version`): not relevant
- [ ] Chart version/commit (`helm list -n posthog`):  `5.1.0`
- [ ] Posthog version: `1.30.0`

tiina303:
@guidoiaquinti

guidoiaquinti:
👋 Hey @Legion2, thank you for your report.

Would be ok if we add a `clickhouse.nodeSelector` variable where you can pass a map? I can't easily change the current `clickhouseOperator.useNodeSelector` without making a breaking change. 

I anyway agree that we should deprecate the current `clickhouseOperator.useNodeSelector` as it is not very useful.

Legion2:
@tiina303 yes I think calling the option `clickhouse.nodeSelector` is the best, becaues it is consistent with the other node selector options of the other components.

guidoiaquinti:
@Legion2 for some reason I can't add you as reviewer but do you mind to 👀 ☝️? It should fix your issue. Thank you and have a good weekend!

PR #212, opened 2021-12-03
Add 'clickhouse.nodeSelector' variable

guidoiaquinti:
## Description
Add a new chart variable `clickhouse.nodeSelector` to customize the `nodeSelector` definition of the ClickHouse installation. Fixes https://github.com/PostHog/charts-clickhouse/issues/210.

Optional: we should deprecate `clickhouseOperator.useNodeSelector`, any user currently using it should simply use:
```
clickhouse.nodeSelector:
  clickhouse: true
```

## Type of change

## How has this been tested?
CI should be passing

## Checklist

### DIFF
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -125,6 +125,9 @@ spec:
                   mountPath: /var/lib/clickhouse
               resources:
 {{ toYaml .Values.clickhouse.resources | indent 16 }}
+              {{- if .Values.clickhouse.nodeSelector }}
+              nodeSelector: {{- .Values.clickhouse.nodeSelector | toYaml | nindent 16 }}
+              {{- end }}
               {{- if .Values.clickhouseOperator.useNodeSelector }}
               nodeSelector:
                 clickhouse: true
--- a/charts/posthog/tests/__snapshot__/clickhouse-instance.yaml.snap
+++ b/charts/posthog/tests/__snapshot__/clickhouse-instance.yaml.snap
@@ -1,0 +1,89 @@
+the manifest should match the snapshot when using default values:
+  1: |
+    apiVersion: clickhouse.altinity.com/v1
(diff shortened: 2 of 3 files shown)


Legion2 (approved):
LGTM

charts/posthog/templates/clickhouse_instance.yaml
```diff
@@ -125,6 +125,9 @@ spec:
                   mountPath: /var/lib/clickhouse
               resources:
 {{ toYaml .Values.clickhouse.resources | indent 16 }}
+              {{- if .Values.clickhouse.nodeSelector }}
+              nodeSelector: {{- .Values.clickhouse.nodeSelector | toYaml | nindent 16 }}
```
Legion2, line 129: this may not have the correct indent, I think this should be at the same indent as the tolerations?

Legion2 (changes requested):

tiina303 (approved):
Could we use sth similar to for encrypted drive too (https://github.com/PostHog/charts-clickhouse/pull/197)?

guidoiaquinti:
> #197

We can but I don't think we should.

charts/posthog/templates/clickhouse_instance.yaml
guidoiaquinti: Looking at [this example](https://github.com/Altinity/clickhouse-operator/issues/450#issuecomment-665728125) I think it should be in the `spec` tree?

Legion2: yes it should be part of the pod spec template and not the container

guidoiaquinti: You are right, the PR is using a bad indentation. I need a pair of 👓

guidoiaquinti:
@Legion2 let me know if this works for you. Thanks! 🙇

Legion2:
@guidoiaquinti I think so, I don't know how to test this PR in our cluster

PR #205, opened 2021-11-26
`clickhouse-operator`: upgrade to version 0.16.1

guidoiaquinti:
## Description
This PR aligns the `clickhouse-operator` resources (templates and CDRs) to the latest release version `0.16.1`. Fixes https://github.com/PostHog/charts-clickhouse/issues/164.

In order to make this process more reproducible I've build a utility tool at `scripts/clickhouse_operator_sync.sh` that should make this process less painful in the future.

Due to the length, I suggest to review this PR commit by commit:

* 1 and 2: those are _preparation commits_ and they are simple `git mv`s. I did this to make the code review easier as git usually interprete a file as new (instead of modified) if ~50% of its content changes. By explicitly moving the files in a commit and then editing them in another, the visualisation should be easer.
* 3: added the utility tool + fixed a minor bug in the `k6` ingestion test I've discovered while working on this
* 4: fixed the unit test template path in order to have them 🟢

--- till this point there are no structural changes to the Helm chart, we simply moved some files around ---

* 5: we run `scripts/clickhouse_operator.sh` to pull `clickhouse-operator@0.16.1`, parse its content and saving it to the appropriate locations.

* 6: we adjust the unit tests to work with the new code
* 7: adjust the codespell config
* 8: before this PR we had a single file for cluster role and cluster role bindings. With this PR we now have 2 distinct files (and we should do the same for the tests)

## Type of change
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## How has this been tested?
**New installs**
1. CI should be passing
1. I've also tested a fresh install locally and everything seems to work as expected

**Update**
1. Switched to the current `main` branch and installed PostHog using it as source:
    ```
    helm upgrade --install --set "cloud=private" -f ./charts/posthog/values.yaml --timeout 10m --create-namespace --namespace posthog posthog ./charts/posthog --wait --wait-for-jobs
    ```

1. Switched to the current branch `clickhouse_operator_upgrade`

1. Updated the `altinity/clickhouse` CRDs:
    ```
    kubectl apply \
        -f charts/posthog/crds/clickhouseinstallations.clickhouse.altinity.com.yaml \
        -f charts/posthog/crds/clickhouseinstallationtemplates.clickhouse.altinity.com.yaml \
        -f charts/posthog/crds/clickhouseoperatorconfigurations.clickhouse.altinity.com.yaml
    ```

[...]

PR #213, opened 2021-12-03
Update Helm chart to support k8s 1.22

guidoiaquinti:
## Description
With https://github.com/PostHog/charts-clickhouse/pull/205 we removed the last roadblock to support Kubernetes version 1.22

This PR enables integration testing against this version and also update the chart metadata. 

## Type of change

## How has this been tested?
CI should be passing.

## Checklist


tiina303 (approved):
LGTM

fuziontech (approved):

hazzadous:
Have kicked off the failing jobs 🤞

hazzadous:
Fyi I added https://github.com/PostHog/charts-clickhouse/pull/220 for debugging

Looks like there might be some issues with ingress: https://github.com/PostHog/charts-clickhouse/runs/4461381717?check_suite_focus=true

guidoiaquinti:
Thank you @hazzadous ! Yeah, I missed that we need to update the Helm dependency `ingress-nginx` to version >= 4:

> Ingress resources evolved over time. They started with apiVersion: extensions/v1beta1, then moved to apiVersion: networking.k8s.io/v1beta1 and more recently to apiVersion: networking.k8s.io/v1.
> 
> Here is how these Ingress versions are supported in Kubernetes: - before Kubernetes 1.19, only v1beta1 Ingress resources are supported - from Kubernetes 1.19 to 1.21, both v1beta1 and v1 Ingress resources are supported - in Kubernetes 1.22 and above, only v1 Ingress resources are supported
> 
> And here is how these Ingress versions are supported in NGINX Ingress Controller: - before version 1.0, only v1beta1 Ingress resources are supported - in version 1.0 and above, only v1 Ingress resources are
> 
> As a result, if you're running Kubernetes 1.19 or later, you should be able to use the latest version of the NGINX Ingress Controller; but if you're using an old version of Kubernetes (1.18 or earlier) you will have to use version 0.X of the NGINX Ingress Controller (e.g. version 0.49).
> 
> The Helm chart of the NGINX Ingress Controller switched to version 1 in version 4 of the chart. In other words, if you're running Kubernetes 1.19 or earlier, you should use version 3.X of the chart (this can be done by adding --version='<4' to the helm install command).
>
> ref: https://kubernetes.github.io/ingress-nginx/deploy/#running-on-kubernetes-versions-older-than-119

guidoiaquinti:
We also need to update `cert-manager` from the current version `1.2.0` to _at least_ `1.5` (see [here](https://cert-manager.io/docs/installation/supported-releases/))

PR #219, opened 2021-12-08
Add ability to expose ClickHouse externally (some clients require this)

fuziontech:
## Description
A customer is requesting that we bring back external access to ClickHouse on their cluster. This allows us to flip a switch to enable that.

### DIFF
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -139,6 +139,15 @@ spec:
             - name: tcp
               port: 9000
           type: {{ .Values.clickhouseOperator.serviceType }}
+        metadata:
+          annotations:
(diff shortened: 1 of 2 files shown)


guidoiaquinti:
> A customer is requesting that we bring back external access to ClickHouse on their cluster. This allows us to flip a switch to enable that.

What is the purpose of exposing CH externally? If it’s for debugging purposes, I think they should simply use `kubectl port-forward`. I’m pushing back on this as I don’t think it’s something we should support in our chart.

guidoiaquinti:
Also, if they need external access they can probably simply override `clickhouseOperator.serviceType`. I'm closing this @fuziontech but feel free to reopen if I've missed anything.

fuziontech:
Yup! The customer already figured it out themselves too. 👍 to closing

Issue #176, opened 2021-11-02
Investigate downtime during deployment


guidoiaquinti:
## Bug description
A customer running on AWS is reporting some downtime during `helm` operations. They are running with 2 replicas for `web`/`events`.

 
## Expected behavior
No downtime expected

## How to reproduce
_"It appears that helm creates two new nodes, and terminates the old ones. But it terminates them before the new pods become ready / healthy. For example I can see this shortly after an upgrade, causing NGINX to show a 503 for all requests."_
```
NAME                                                READY   STATUS    RESTARTS   AGE
posthog-web-c4bc4b487-4msjb                         0/1     Running   0          54s
posthog-web-c4bc4b487-cxzhn                         0/1     Running   0          54s
```
## Environment

- [X] Deployment platform (gcp/aws/...): AWS (possibly everywhere)
- [X] Chart version/commit: latest
- [X] Posthog version: latest

## Additional context
* See [private Slack channel](https://posthogusers.slack.com/archives/C02EAFFA11S/p1635866041034200)

Legion2:
I think the downtime is related to the way the deployments are added via Helm, all deployments are annotated as helm hooks, so helm will create and delete them as part of this lifecycle.
https://github.com/PostHog/charts-clickhouse/blob/4273cd1e9bc982cc1bc08642f0bcad759deeec13/charts/posthog/templates/web-deployment.yaml#L11-L15
I think this is a mistake, that the deployments are configured as hooks.

guidoiaquinti:
> I think the downtime is related to the way the deployments are added via Helm, all deployments are annotated as helm hooks, so helm will create and delete them as part of this lifecycle.
> 
> https://github.com/PostHog/charts-clickhouse/blob/4273cd1e9bc982cc1bc08642f0bcad759deeec13/charts/posthog/templates/web-deployment.yaml#L11-L15
> 
> 
> I think this is a mistake, that the deployments are configured as hooks.

You are on point! [Here's](https://github.com/PostHog/charts-clickhouse/pull/179) an attempt to remove them but it's currently blocked by some issues in `posthog/posthog`

PR #221, opened 2021-12-09
Dependency: `ingress-nginx` update from '3.25.0' to '4.0.13'

guidoiaquinti:
## Description
This PR updates the `ingress-nginx` dependency chart from `3.25.0` to `4.0.13`.

Looking at [the upstream changelog](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/CHANGELOG.md) the only breaking change seems to be the deprecation of Kubernetes versions < v1.19 (by dropping the support for Ingress Object in `networking.k8s.io/v1beta`). This should be a no-op on our end as we already dropped support for it before. Everything else should be retro-compatible.

## Type of change

## How has this been tested?
1. CI should be passing ✅ 

2. I've also setup a fresh new installation on DO using the `main` branch. Then switched to this one and ran:
    ```
    helm upgrade --install -f /tmp/values.yaml --timeout 20m --create-namespace --namespace posthog posthog ./charts/posthog --wait --wait-for-jobs
    ```
    the ingress controller pod gets terminated and updated as expected. There's technically a brief downtime between pod termination and re-spawn but it was so small that I wasn't able to track it (looks like < 1s). Anyway, we experience already way longer downtimes due to [this](https://github.com/PostHog/charts-clickhouse/issues/176) so I don't think we should consider this an issue.

## Checklist


tiina303 (approved):
🚢

PR #222, opened 2021-12-09
Deprecate `clickhouseOperator.useNodeSelector`

guidoiaquinti:
## Description
With https://github.com/PostHog/charts-clickhouse/pull/212 we added a new `clickhouse.nodeSelector` chart variable to customize the `nodeSelector` definition of the ClickHouse installation.

This PR removes the now deprecated `clickhouseOperator.useNodeSelector` chart value. Looking at the how `altinity/clickhouse-operator` works, I doubt this value in our repo had ever worked as expected. 

In any case, to keep any eventual retro compatibility `clickhouseOperator.useNodeSelector: true` can be modified as:
```
clickhouse:
  nodeSelector:
    clickhouse: "true"
```

## Type of change

## How has this been tested?
CI is passing

## Checklist

### DIFF
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -110,11 +110,6 @@ spec:
                   mountPath: /var/lib/clickhouse
               {{- end }}
 
-              {{- if .Values.clickhouse.useNodeSelector }}
-              nodeSelector:
-                clickhouse: true
-              {{- end }}
-
               {{- if .Values.clickhouse.resources }}
               resources: {{ toYaml .Values.clickhouse.resources | nindent 16 }}
               {{- end }}
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -552,9 +552,6 @@ clickhouse:
   # -- Service Type: LoadBalancer (allows external access) or NodePort (more secure, no extra cost)
   serviceType: NodePort
 
-  # -- If enabled, operator will prefer k8s nodes with tag `clickhouse:true`
-  useNodeSelector: false
-
   persistence:
     # -- Enable data persistence using PVC.
     enabled: true
--- a/charts/posthog/templates/_snippet-error-on-removed-values.tpl
+++ b/charts/posthog/templates/_snippet-error-on-removed-values.tpl
@@ -20,6 +20,9 @@
 {{- if ne .Values.postgresql.postgresqlUsername "postgres" }}
     {{- required (printf (include "snippet.error-on-removed-values-template" .) "postgresql.postgresqlUsername has been removed" "https://posthog.com/docs/self-host/deploy/upgrade-notes#upgrading-from-13xx") nil -}}
 {{- end -}}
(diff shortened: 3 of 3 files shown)


tiina303:
LGTM thanks

tiina303:
could you also add the new nodeSelector to values with a comment explaining it, so it's more easily discoverable for us & users

tiina303:
related https://github.com/PostHog/posthog.com/pull/2671

fuziontech:
I believe Vendasta is using this so I'll check with them to make sure they have new settings in place.

tiina303:
can we move forward with this? Yesterday the useNodeSelector setting didn't work on a self-hosted install and rather than investigate further I think we should just move ahead and deprecate as has been the plan anyway.

guidoiaquinti:
> can we move forward with this? Yesterday the useNodeSelector setting didn't work on a self-hosted install and rather than investigate further I think we should just move ahead and deprecate (to avoid confusion) as has been the plan anyway.

Yes we can and I think we just confirmed what I thought:

> Looking at the how altinity/clickhouse-operator works, I doubt this value in our repo had ever worked as expected.

If this never worked, I'm 👍 to also don't make this a breaking change (as it never worked before)

charts/posthog/templates/clickhouse_instance.yaml
```diff
@@ -110,11 +110,6 @@ spec:
                   mountPath: /var/lib/clickhouse
               {{- end }}
 
-              {{- if .Values.clickhouse.useNodeSelector }}
-              nodeSelector:
```
hazzadous, line 114: yeah looks like this should be at the pod spec context and not the container context? As per https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/

hazzadous (approved):
Looks good, as for update it seems there's no harm in making this a non-breaking change but we should obviously have somewhere in the release notes that they should use `nodeSelector` instead.

hazzadous:
> As we usually write release notes for breaking changes only, what do you think about simply updating the docs?

Yeah I think that’s ok, if a user is having issues with it hopefully they’ll do some searches and see that it’s deprecated from GitHub issues and that it never worked, from the pr

guidoiaquinti:
We can also add a deprecation notice here in Helm. Good suggestion

hazzadous (approved):

tiina303:
how about we update the already existing release note about this https://posthog.com/docs/self-host/deploy/upgrade-notes#upgrading-from-9xx

charts/posthog/values.yaml
```diff
@@ -552,9 +552,6 @@ clickhouse:
   # -- Service Type: LoadBalancer (allows external access) or NodePort (more secure, no extra cost)
   serviceType: NodePort
 
-  # -- If enabled, operator will prefer k8s nodes with tag `clickhouse:true`
-  useNodeSelector: false
```
tiina303, line 556: can we get `nodeSelector` documented in values it currently isn't (i.e. isn't in ALL_VALUES)

tiina303 (approved):

Issue #215, opened 2021-12-06
Allow to specify a custom storage class for clickhouse instance


Legion2:
Currently the clickhouse instance can either be configured with an external PVC or a volumeClaimTemplate. However, the storage class of the volumeClaimTemplate can not be customized. 

https://github.com/PostHog/charts-clickhouse/blob/9000e659a96f42869315552e668cf947701fbb60/charts/posthog/templates/clickhouse_instance.yaml#L146-L150

### Proposed change
The `storageClassName` should be an optional parameter of the helm chart, if not set by the user the value is choosen based on the cloud provider. Please, also consider to use the `global.storageClass` value as default which is supported by all bitnami helm charts to globaly specify the default storage class for all persistent volumes if not specified otherwise.

### Alternative options
Workaround for setting a custom storageClass is to use the external PVC option, however then replication cannot be used.

Legion2:
If a custom storage class is used the generation of the storage classes should be disabled.

Usecase: We want to use aws gp3 storage with the new EBS CSI driver and therefore created a custom storage class.

guidoiaquinti:
Closed via #224

Issue #214, opened 2021-12-04
Allow option for PVC’s to be encrypted (AWS)


joesaunderson:
### Proposed change

We use AWS as our cloud provider and our SRE is reporting the PVC's created by the PostHog helm chart are not conforming to our AWS Control Tower policies, this is because they are not encrypted.

Further to this, when turning on the 'Default volumes to encrypted' setting in AWS, the whole helm chart fails to install altogether.

It would be good to be able to specify the PVC as encrypted. 

### Alternative options

Don't encrypt the drives and have a non conforming stack based on our Control tower policies.

guidoiaquinti:
Closed via #224

Issue #120, opened 2021-09-14
Automated test for deploying to a different namespace


tiina303:
## Is your feature request related to a problem?

We have broken chart deployment to namespaces other than "posthog" various times.

## Describe the solution you'd like

Add a test that tries to deploy with different namespace name & check everything came up.

## Describe alternatives you've considered

as is; use a different namespace during release testing.

sdenardi:
Hi there, we recently came upon this issue while setting up our self-hosted deployments to AWS. We want to deploy a Sandbox environment alongside Production within the same k8s Cluster, matching Prod as closely as possible so we can test out any changes or upgrades.

We attempted to deploy both Sandbox and Production first using different namespaces, but ran into the issues that I assume were the impetus for this GH Issue. We then tried deploying the charts with the same release name, but into different namespaces, but ran into errors that resources already exist.

Any movement on a fix for this would be greatly appreciated.

PH Version: `1.35.0`
k8s Version: `1.21.9`

tiina303:
I would encourage different k8s clusters for sandbox and test. To avoid potential problems of sandbox taking away resources from prod. 
Regarding the problems you ran into could you provide more details?

sdenardi:
When we tried deploying the helm chart into a namespace named `posthog-prod`, the Kafka pod entered into an endless crash loop while trying to connect to ClickHouse.

tiina303:
> When we tried deploying the helm chart into a namespace named posthog-prod, the Kafka pod entered into an endless crash loop while trying to connect to ClickHouse.

You need to update https://github.com/PostHog/charts-clickhouse/blob/main/charts/posthog/values.yaml#L575 Unfortunately we haven't found a way to have that work natively.

That said I'd like to repeat my suggestion to use different clusters as there can be various problems now or in the future as running two instances in the same cluster it's not something we specifically support. This issue was mainly about having a single instance of Posthog running just in a different namespace.

PR #224, opened 2021-12-10
Make StorageClass installation optional in AWS/GCP

guidoiaquinti:
## Description
1️⃣  Till this PR we've been including additional `StorageClass` definition into our chart only when using AWS and GCP. As extensively already discussed, this should be considered an anti-pattern as storage class definitions are global resources that only administrator should be able to manage.

Our chart should simply rely on the cluster default storage class and/or give our users the ability to customise which `StorageClass` to use.

2️⃣ This PR renames the `clickhouseOperator` values to `clickhouse` as they were never used to configured the operator:

* `clickhouseOperator.enabled` -> `clickhouse.enabled`
* `clickhouseOperator.namespace` -> `clickhouse.namespace`
* `clickhouseOperator.storage` -> `clickhouse.persistence.size`
* `clickhouseOperator.useNodeSelector` -> `clickhouse.useNodeSelector`
* `clickhouseOperator.serviceType` -> `clickhouse.serviceType`

we also renamed `clickhouse.persistentVolumeClaim` to `clickhouse.persistence.existingClaim` to align the naming with the  bitnami charts.

Related issues:
* https://github.com/PostHog/charts-clickhouse/issues/174
* https://github.com/PostHog/charts-clickhouse/issues/214
* https://github.com/PostHog/charts-clickhouse/issues/215

## Type of change

## How has this been tested?
- CI is ✅ 
- Verified that if a AWS/GCP user doesn't read the release note and runs an upgrade with the default `installCustomStorageClass: false` the custom `StorageClass` is removed but the underlying volumes don't get deleted. 

## Checklist

### DIFF
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -1,63 +1,34 @@
-{{- if .Values.clickhouseOperator.enabled }}
-{{- if (eq (include "clickhouse.externalVolume" .) "false" ) }}
-{{ if (eq (.Values.cloud | toString) "gcp" )}}
-apiVersion: storage.k8s.io/v1
-kind: StorageClass
-metadata:
-  name: gce-resizable
-provisioner: kubernetes.io/gce-pd
-parameters:
-  type: pd-standard
-  fstype: ext4
-  replication-type: none
-reclaimPolicy: Retain
-#volumeBindingMode: Immediate
-allowVolumeExpansion: true
-{{- else if (eq (.Values.cloud | toString) "aws") }}
-#
-# AWS resizable disk example
-#
-apiVersion: storage.k8s.io/v1
-kind: StorageClass
-metadata:
-  name: gp2-resizable
-provisioner: kubernetes.io/aws-ebs
-parameters:
-  type: gp2
-reclaimPolicy: Retain
-#volumeBindingMode: Immediate
-allowVolumeExpansion: true
-{{ end }}
-{{- end }}
----
+{{- if .Values.clickhouse.enabled }}
 apiVersion: "clickhouse.altinity.com/v1"
 kind: "ClickHouseInstallation"
 metadata:
   name: "posthog"
 spec:
-  defaults:
-    templates:
-      {{- if (eq (include "clickhouse.externalVolume" .) "false" ) }}
-      dataVolumeClaimTemplate: data-volumeclaim-template
-      {{- end }}
-      serviceTemplate: chi-service-template
   configuration:
     users:
       admin/password: {{ .Values.clickhouse.password }}
       admin/networks/ip: "0.0.0.0/0"
       admin/profile: default
       admin/quota: default
+
     profiles:
       default/allow_experimental_window_functions: "1"
+
     clusters:
       - name: {{ .Values.clickhouse.database | quote }}
         templates:
-          podTemplate: pod-template-with-volumes
+          podTemplate: pod-template
+          serviceTemplate: service-template
+          {{- if and (.Values.clickhouse.persistence.enabled) (not .Values.clickhouse.persistence.existingClaim) }}
+          dataVolumeClaimTemplate: data-volumeclaim-template
+          {{- end }}
         layout:
           shardsCount: 1
           replicasCount: 1
+
     settings:
       format_schema_path: /etc/clickhouse-server/config.d/
+
     files:
       events.proto: |
         syntax = "proto3";
@@ -71,6 +42,7 @@ spec:
           string created_at = 7;
           string elements_chain = 8;
         }
+
     zookeeper:
       nodes:
       {{- if .Values.clickhouse.externalZookeeper }}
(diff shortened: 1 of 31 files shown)


charts/posthog/templates/clickhouse_instance.yaml
```diff
@@ -1,35 +1,4 @@
 {{- if .Values.clickhouseOperator.enabled }}
-{{- if (eq (include "clickhouse.externalVolume" .) "false" ) }}
```
guidoiaquinti, line 2: Simply moved to a dedicated file

hazzadous:
> Verified that if a AWS/GCP user doesn't read the release note and runs an upgrade with the default installCustomStorageClass: false the custom StorageClass is not removed and/or volumes don't get deleted: TODO

@guidoiaquinti I'll test this on staging environment, by hand will do for now?

guidoiaquinti:
> @guidoiaquinti I'll test this on staging environment, by hand will do for now?

Yes I think it's good. Fresh install from `main` then upgrade from `remove_custom_storage_classes` (this branch). I'm going to do the same but for GCP.

guidoiaquinti:
Draft release notes https://github.com/PostHog/posthog.com/pull/2757

hazzadous:
I've run through and it looks like it:

 1. removes the storageclass
 2. keeps the pv, and doesn't delete this after pod restart

https://gist.github.com/hazzadous/f078934a2367f3433c09c8d7a5c0ba13

ci/kubetest/test_clickhouse_persistence_disabled.py
```diff
@@ -0,0 +1,30 @@
+import pytest
```
hazzadous, line 1: Now that we have actual kubetest, could we remove the `test_something.py`?

ci/kubetest/test_clickhouse_persistence_disabled.py
```diff
@@ -0,0 +1,30 @@
...
+
+from utils import cleanup_k8s, get_clickhouse_statefulset_spec, helm_install, wait_for_pods_to_be_ready
+
+HELM_INSTALL_CMD = """
+helm upgrade \
+    --install \
+    -f ../../ci/values/kubetest/test_clickhouse_persistence_disabled.yaml \
+    --timeout 30m \
+    --create-namespace \
+    --namespace posthog \
```
hazzadous, line 11: non-blocking

Worth putting these test cases into separate namespaces for better isolation?

ci/kubetest/test_clickhouse_persistence_disabled.py
```diff
@@ -0,0 +1,30 @@
...
+    helm_install(HELM_INSTALL_CMD)
+    wait_for_pods_to_be_ready(kube)
+
+
+def test_volume_claim(setup, kube):
+    statefulset_spec = get_clickhouse_statefulset_spec(kube)
+
+    # Verify the spec.volumes configuration
+    volumes = statefulset_spec.template.spec.volumes
+    assert all(volume.config_map is not None for volume in volumes), "All the spec.volumes should be of type config_map"
```
hazzadous, line 30: non-blocking

Is `volume.config_map is not None` saying that persistence is disabled?

ci/kubetest/test_clickhouse_persistence_disabled.py
guidoiaquinti: Fair point and I have the `git rm` in another branch

ci/kubetest/test_clickhouse_persistence_disabled.py
guidoiaquinti: It could be nice 😅 the problem is that I don't think our chart works in a namespace != posthog (see [here](https://github.com/PostHog/charts-clickhouse/issues/120))

hazzadous (approved):
I think it looks great although it's touching some high risk areas so would love other eyes

Not important non-blocking point, it would be nice to have e.g. the `clickhouseOperator` name change as a separate commit.

ci/kubetest/test_clickhouse_persistence_disabled.py
hazzadous: Ah ok cool, in my test I did use non-posthog namespace but yeah as long as the tests are passing and checking the right things I don't think it matters.

ci/kubetest/test_clickhouse_persistence_disabled.py
guidoiaquinti: It means the only volume(s) attached to the container are of type [configmap](https://kubernetes.io/docs/concepts/configuration/configmap/). We could also flip the logic if you think it can be more clear

tiina303 (approved):
lgtm just a consideration for nodeselector

ci/kubetest/test_clickhouse_persistence_disabled.py
hazzadous: Yeah I'm not totally up to speed here but I think validating that there is no PVC would be more clear?

Anyway, it's not a big deal

charts/posthog/templates/clickhouse_instance.yaml
```diff
@@ -112,49 +91,57 @@ spec:
...
                   mountPath: /var/lib/clickhouse
-              resources:
-{{ toYaml .Values.clickhouse.resources | indent 16 }}
-              {{- if .Values.clickhouseOperator.useNodeSelector }}
+              {{- end }}
+
+              {{- if .Values.clickhouse.useNodeSelector }}
               nodeSelector:
                 clickhouse: true
               {{- end }}
```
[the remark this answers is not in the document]
guidoiaquinti: I think we should do the deprecation via #222 as there are already too many changes in this one.

PR #226, opened 2021-12-13
Add celery worker cpu request

yakkomajuri:
## Description

**Friendly reminder that I'm way out of my depth with Helm charts.**

Essentially, as part of the special migrations work, I'd like to bump Celery specs a bit.

There are 2 ways to do this:

- Set min replicas to 2
- Bump CPU request

Here I'm going with a CPU request bump as Celery will run _n_ worker processes where _n == n cores_. Our default deployments seem to be giving us 2 cores, and adding an extra one ensures Celery can have the same throughput while we hijack a worker for special migrations.

I guess this is a breaking change? Maybe the replicas approach is better?

Just sparking discussion here. I'd rather have a sensible default than write on the release notes that people should bump celery replicas on upgrades.

## Type of change

## How has this been tested?

It hasn't

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -169,7 +169,8 @@ worker:
   replicacount: 1
   # -- Resource limits for workers
   resources:
-    {}
+    requests:
+      cpu: 3000m
     # limits:
     #   cpu: 300m
     #   memory: 500Mi


guidoiaquinti:
👋 Hey Yakko, welcome back in this repo!

I'm 100% into adding resource limits (see https://github.com/PostHog/charts-clickhouse/pull/159) but as we didn't do this before we should consider this a breaking change. See also the considerations already done in the other PR.

yakkomajuri:
@guidoiaquinti what would we need to get this over the line? can we batch this with the next breaking release?

guidoiaquinti:
> @guidoiaquinti what would we need to get this over the line? can we batch this with the next breaking release?

I think we should make this change as part of #159.

hazzadous:
I going to close this to reduce the noise in this repo, if it's really important feel free to open again.

PR #229, opened 2021-12-15
Fix GCP & bump github action timeout, etc...

guidoiaquinti:
## Description
1. Fix GCP e2e test
1. Swap primary DNS used by `cert-manager` pods from CloudFlare to Google.
1. Use smaller disks in CI

## Why
1. I've just figured out that [Network endpoint groups](https://cloud.google.com/load-balancing/docs/negs) aka NEGs, are not deleted when the GKE cluster is deleted. 5 days ago we ended up saturating the # of NEG allowed in an account and the CI action started failing. I've added some steps in the CI action to perform this cleanup.
1. From some manual tests I did it seems DNS propagation is faster
1. Let's provision only what we need

## Type of change

## How has this been tested?
CI is passing

## Checklist


tiina303 (approved):
🚢

.github/workflows/test-amazon-web-services-install.yaml
```diff
@@ -77,6 +77,7 @@ jobs:
           --nodes 3
 
     - name: Install PostHog using the Helm chart
+      id: helm_install
```
hazzadous, line 80: non-blocking

Is this actually used at all?

guidoiaquinti: Not yet but I find useful to keep the steps identical across e2e tests so maybe we will be able to extract and move them out to reduce copy paste (not a pri for now)

.github/workflows/test-google-cloud-platform-install.yaml
```diff
@@ -175,6 +208,28 @@ jobs:
...
+            do
+              gcloud compute network-endpoint-groups delete \
+                  --project ${{ secrets.GCP_PROJECT_ID }} \
+                  --zone "$NEG_ZONE" \
+                  --quiet \
+                  "$NEG_NAME"
+            done
+        }
+
+        delete_neg "${{ steps.vars.outputs.posthog_events_neg_name }}" "${{ steps.vars.outputs.posthog_events_neg_zones }}"
```
hazzadous, line 230: non-blocking

I guess these can be single quotes as context substitution happens at workflow gen time not at run time?

hazzadous (approved):
Looks alright

I haven't played with GCP too much, but is there a clean way of

1. create new project
2. setup everything and run tests
3. destroy project and all resources?

.github/workflows/test-google-cloud-platform-install.yaml
guidoiaquinti: Fair point, it's just habit :D

guidoiaquinti:
> * create new project
> * setup everything and run tests
> * destroy project and all resources?

Unfortunately not (as for each project you also need to setup billings etc..).

.github/workflows/test-google-cloud-platform-install.yaml
```diff
@@ -175,9 +208,27 @@ jobs:
           --quiet \
           ${{ steps.vars.outputs.k8s_cluster_name }}
 
-    # TODO: fix this!
-    # When a GKE service is deleted, the associated NEG will not be garbage collected if the NEG is still referenced by a backend service. Dereference the NEG from the backend service to allow NEG deletion.
-    # When a cluster is deleted, standalone NEGs are not deleted.
+    - name: Delete the associated NEG (Network Endpoint Groups)
+      if: ${{ always() && steps.helm_install.outcome == 'success' && steps.fetch_neg.outcome == 'success' }}
```
tiina303, line 212: helm install might have failed, but we still might have web or events up, so maybe it would be better to make this dependent on if we were able to fetch negs and get some values & always try to run the fetch above too

PR #233, opened 2021-12-21
Fix typo in `charts/posthog/Chart.lock` :facepalm:

guidoiaquinti:
## Description
This typo has been introduced via #1 and it's present since the beginning of this repo. Everywhere in the codebase we use `kafka.enabled` while in the logic to include it or not we used `kafka.enable`. This mean we are always installing Kafka also when passing the `kafka.enabled` override.

I've discovered this bug while working on more integration tests I'll probably push after the holiday break.  

## Type of change

## How has this been tested?
CI is passing + local testing

## Checklist


marcushyett-ph (approved):
LGTM

PR #234, opened 2021-12-23
K8s: move supported version to >=1.20 <= 1.23

guidoiaquinti:
## Description
* drop support for k8s version 1.19 as it is end of life since 2021-10-28
* add support for k8s version 1.23 [announced on Dec 7th](https://kubernetes.io/blog/2021/12/07/kubernetes-1-23-release-announcement/)

~~Note: CI for 1.23 is still failing as there isn't yet k3s support available.~~

## Type of change

## How has this been tested?
CI should be passing ✅ 

## Checklist

### DIFF
--- a/.github/workflows/test-helm-chart.yaml
+++ b/.github/workflows/test-helm-chart.yaml
@@ -21,13 +21,13 @@ jobs:
         # specified below, these parameters have no meaning on their own and
         # gain meaning on how job steps use them.
         include:
+          - k3s-channel: v1.23
+            test: install
           - k3s-channel: v1.22
             test: install
           - k3s-channel: v1.21
             test: install
           - k3s-channel: v1.20
-            test: install
-          - k3s-channel: v1.19
             test: install
 
           # We run an upgrade test where we first install an already released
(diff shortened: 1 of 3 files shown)


.github/workflows/test-helm-chart.yaml
```diff
@@ -21,22 +21,22 @@ jobs:
...
-            test: install
 
           # We run an upgrade test where we first install an already released
           # Helm chart version and then upgrade to the version we are now
           # testing. We test upgrading from the latest main version.
           #
           # It can be very useful to see the "Helm diff" step's output from the
           # latest dev version.
-          - k3s-channel: v1.19
+          - k3s-channel: v1.21
```
hazzadous, line 39: What's the reason for using v1.21 here?

hazzadous (approved):
Looks good to me.

I'm guessing the upgrade k8s version used is arbitrary. non blocking.

.github/workflows/test-helm-chart.yaml
guidoiaquinti: We are currently running the upgrade only on a single platform (with this change 1.21). I've picked that because it's one of the versions most used by our users.

PR #236, opened 2022-01-04
Change workflow config to be less aggressive

guidoiaquinti:
## Description
Change the workflow config to only run when it is really necessary. We currently run all the workflows too often. For the e2e test (as they have a cost) let's reduce this to manual execution.  

## Type of change

## How has this been tested?
CI should be passing

## Checklist

### DIFF
--- a/.github/workflows/test-lint-helm.yaml
+++ b/.github/workflows/test-lint-helm.yaml
@@ -4,7 +4,6 @@ on:
   workflow_dispatch:
   push:
     paths:
-      - ci/**
       - charts/**
       - .github/workflows/test-lint-helm.yaml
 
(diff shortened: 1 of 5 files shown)


.github/workflows/test-lint-helm.yaml
```diff
@@ -4,7 +4,6 @@ on:
   workflow_dispatch:
   push:
     paths:
-      - ci/**
```
guidoiaquinti, line 7: Helm lint doesn't use anything in the `ci/` folder

.github/workflows/test-unit-helm.yaml
```diff
@@ -4,7 +4,6 @@ on:
   workflow_dispatch:
   push:
     paths:
-      - ci/**
```
guidoiaquinti, line 7: Helm unittest doesn't use anything in the `ci/` folder

marcushyett-ph (approved):

PR #237, opened 2022-01-05
CI - Fix helm-diff

guidoiaquinti:
## Description
The `helm-diff` plugin we use in CI is broken since ~15h due to a regression introduced upstream. Let's pin to the latest stable version 3.1.3 till it doesn't get fixed.

See upstream issues:
- https://github.com/databus23/helm-diff/issues/322
- https://github.com/databus23/helm-diff/issues/324

## Type of change

## How has this been tested?
CI should be passing

## Checklist

### DIFF
--- a/.github/workflows/test-helm-chart.yaml
+++ b/.github/workflows/test-helm-chart.yaml
@@ -1,7 +1,4 @@
-# This is a GitHub workflow defining a set of jobs with a set of steps.
-# ref: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
-#
-name: Test chart
+name: e2e - k3s
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-helm-chart.yaml
```diff
@@ -78,10 +75,11 @@ jobs:
 
       - name: "(Upgrade) Await chart install"
         if: matrix.test == 'upgrade'
-        uses: jupyterhub/action-k8s-await-workloads@v1
+        uses: jupyterhub/action-k8s-await-workloads@main
         with:
-          timeout: 900
-          max-restarts: 15
+          namespace: "posthog"
```
guidoiaquinti, line 80: Not strictly related to this PR but we were watching the wrong namespace

.github/workflows/test-helm-chart.yaml
```diff
@@ -69,7 +66,7 @@ jobs:
       - name: "(Upgrade) Install helm diff"
         if: matrix.test == 'upgrade'
         run: |
-          helm plugin install https://github.com/databus23/helm-diff
+          helm plugin install https://github.com/databus23/helm-diff --version 3.1.3
```
tiina303, line 69: can we do sth like `--version stable` do they have a tag?

tiina303 (approved):

.github/workflows/test-helm-chart.yaml
guidoiaquinti: I don't think they do but I'm ok to go back on `master`/`main` once they fix

PR #239, opened 2022-01-07
CI - add kubetest

guidoiaquinti:
## Description
This PR adds a new type of test suite using `kubetest` (you can find more info about it in the `README.md` that I've added to this PR).

This is a subset of https://github.com/PostHog/charts-clickhouse/pull/224 that I'm splitting into multiple parts.

## Type of change

## How has this been tested?
CI should be passing

## Checklist

### DIFF
--- a/ci/kubetest/README.md
+++ b/ci/kubetest/README.md
@@ -1,6 +1,5 @@
 # Kubetest
 This directory contains the integration tests for our Helm chart. We use [`kubetest`](https://kubetest.readthedocs.io/en/latest/) that is a `pytest` plugin that makes it easier to write integration tests on Kubernetes.
-x
 
 ### How-to
 1. `python3 -m venv .venv`
--- a/ci/kubetest/README.md
+++ b/ci/kubetest/README.md
@@ -1,0 +1,18 @@
+# Kubetest
+This directory contains the integration tests for our Helm chart. We use [`kubetest`](https://kubetest.readthedocs.io/en/latest/) that is a `pytest` plugin that makes it easier to write integration tests on Kubernetes.
+x
+
+### How-to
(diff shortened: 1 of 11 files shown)


ci/kubetest/test_placeholder.py
```diff
@@ -0,0 +1,10 @@
+import pytest
```
guidoiaquinti, line 1: This will be removed when real test will be added

ci/kubetest/README.md
```diff
@@ -0,0 +1,18 @@
+# Kubetest
+This directory contains the integration tests for our Helm chart. We use [`kubetest`](https://kubetest.readthedocs.io/en/latest/) that is a `pytest` plugin that makes it easier to write integration tests on Kubernetes.
+x
```
hazzadous, line 3:
```suggestion
```

hazzadous (approved):
Looks good

Non-blocking: I might be missing some context, but what does kubetest give us that we'd not handling with e.g. helm test, or just using `kubectl`?

guidoiaquinti:
> Non-blocking: I might be missing some context, but what does kubetest give us that we're not able to handle with e.g. helm test, or just using `kubectl`?

Good question! I need to add more documentation about the difference but the tldr is the following:

- `helm lint`: lint test suite to verify if the template can be rendered
- `helm unittest`: unit test suite to verify if `rendered template == expected output`
- `kubetest`: integration test suite to verify if _applying the expected output_ against a k8s target cluster gives us also _the expected infrastructure_ (e.g. are the disk really encrypted? Is this environment variable set as expected?)

To see some `kubetest` test scenarios see https://github.com/PostHog/charts-clickhouse/pull/224/files

PR #240, opened 2022-01-07
Add 'enabled' flag for all services

guidoiaquinti:
## Description
This PR introduces few new Helm values to make the installation of all components optional:

- `pgbouncer.enabled`
- `migrate.enabled`
- `events.enabled`
- `web.enabled`
- `worker.enabled`
- `plugins.enabled`

all the above defaults are `true` so this change is technically a no-op. I've also took the opportunity to add:
 
- `events.hpa.enabled`
- `events.hpa.cputhreshold`
- `events.hpa.minpods`
- `events.hpa.maxpods`
- `events.hpa.replicacount`

to be able to control the HPA of the events deployment. 

This PR will be useful for test suite only verifying a single component (e.g. why should I install PostgreSQL if my test suite is targeting ClickHouse?)  

## Type of change

## How has this been tested?
CI should be passing

## Checklist

### DIFF
--- a/charts/posthog/templates/events-hpa.yaml
+++ b/charts/posthog/templates/events-hpa.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.web.hpa.enabled -}}
+{{- if and .Values.events.enabled .Values.events.hpa.enabled -}}
 apiVersion: autoscaling/v1
 kind: HorizontalPodAutoscaler
 metadata:
@@ -13,7 +13,7 @@ spec:
     kind: Deployment
     apiVersion: apps/v1
     name: {{ template "posthog.fullname" . }}-events
-  minReplicas: {{ .Values.web.hpa.minpods }}
-  maxReplicas: {{ .Values.web.hpa.maxpods }}
-  targetCPUUtilizationPercentage: {{ .Values.web.hpa.cputhreshold }}
+  minReplicas: {{ .Values.events.hpa.minpods }}
+  maxReplicas: {{ .Values.events.hpa.maxpods }}
+  targetCPUUtilizationPercentage: {{ .Values.events.hpa.cputhreshold }}
 {{- end }}
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -39,6 +39,8 @@ env:
 
 # -- PgBouncer setup
 pgbouncer:
+   # -- Whether to install PGBouncer or not
+  enabled: true
   hpa:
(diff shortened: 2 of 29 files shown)


charts/posthog/templates/events-hpa.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.web.hpa.enabled -}}
+{{- if and .Values.events.enabled .Values.events.hpa.enabled -}}
```
guidoiaquinti, line 1: Events and web are two different services running the same code. I think it make sense to split them up completely (for real)

charts/posthog/values.yaml
```diff
@@ -59,7 +61,30 @@ pgbouncer:
...
+    # -- This experimental
+    enabled: false
+    # -- CPU threshold percent for the events stack
+    cputhreshold: 60
+    # -- Min pods for the events stack
+    minpods: 1
+    # -- Max pods for the events stack
+    maxpods: 10
+  # -- Amount of events pods to run. Ignored if hpa is used
+  replicacount: 1
```
hazzadous, line 83: non-blocking

I can't see where this is being used

charts/posthog/templates/events-hpa.yaml
hazzadous: non-blocking

What does splitting them up mean btw

hazzadous (approved):
Looks good

I would have thought the driver for this would be more about allowing users to substitute in their own service implementations and disable the inbuilt ones rather that "test suite only verifying a single component" though?

charts/posthog/templates/events-hpa.yaml
guidoiaquinti: Before this change we were using `web.hpa` values to configure the HPA of the `events` deployment. As those are two distinct services (that can scale in different ways) I think we should separate them.

charts/posthog/values.yaml
guidoiaquinti: Good catch, this is actually a "bug" or a missed implementation (see next commit)

guidoiaquinti:
> I would have thought the driver for this would be more about allowing users to substitute in their own service implementations and disable the inbuilt ones rather that "test suite only verifying a single component" though?

This can be another use case of this PR. I didn't advertise it like that as I'm not sure if Helm install would even work if some components are disabled and some other not (technically it should but I didn't test it)

hazzadous (approved):

fuziontech:
This is super helpful for helm chart on cloud. Nice work! 📈

PR #243, opened 2022-01-10
Update kafka default partitions number per topic

yakkomajuri:
## Description

Partitions per kafka topic is currently set to the default of 1. This makes scaling horizontally services like the plugin servers a problem, as additional servers will sit and do nothing since one plugin server is already handling the only partition available. 

## Type of change

## How has this been tested?

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -452,6 +452,8 @@ kafka:
   logRetentionBytes: _15_000_000_000
   # -- The minimum age of a log file to be eligible for deletion due to age
   logRetentionHours: 24
+  # -- The default number of partitions per topic	
+  numPartitions: 12
   zookeeper:
     # -- Install zookeeper on kubernetes
     enabled: false


fuziontech (approved):
lgtm 🚢 it

tiina303 (commented):
Awesome find 👍  I predict we need to tune the retention policy too though.

charts/posthog/values.yaml
```diff
@@ -452,6 +452,8 @@ kafka:
   logRetentionBytes: _15_000_000_000
```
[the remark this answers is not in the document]
fuziontech: Good catch https://kafka.apache.org/documentation/#topicconfigs_retention.bytes

This will need to be adjusted

fuziontech (changes requested):
https://github.com/PostHog/charts-clickhouse/pull/243/files#r781464220 needs to be addressed first (because of issues with kafka running out of space potentially)

We should make sure that this is always set to 70% of available storage meaning more precisely we need to set the formula to be:
`(Size of kafka disk * 0.7)/(numPartitions * 2)` - This is ok for now because the main 2 topics of size is the in flight events to be processed by the plugin server and the events to be ingested by ClickHouse.

Does this make sense?

tiina303:
Here's a cool article as well https://medium.com/event-driven-utopia/understanding-kafka-topic-partitions-ae40f80552e8

My understanding of the world: we have T topics, each of which can have P_T partitions, so to calculate the retention_bytes we'd actually want to `(Size of kafka disk * 0.7)/(sum over all T (P_T))`

Note that we have more than just two topics. Additionally I just tried to update my test DO cluster to have `kafka.numPartitions = 7` and it didn't work, specifically in the kafka pod I saw (note that I did nuke the Kafka pvc, sts & pod before running the upgrade to be sure it applies)
```
/opt/bitnami/kafka/bin/kafka-topics.sh --bootstrap-server localhost:9092 --describe
Topic: clickhouse_events_proto	PartitionCount: 1	ReplicationFactor: 1	Configs: flush.ms=1000,segment.bytes=1073741824,flush.messages=10000,max.message.bytes=1000012,retention.bytes=15000000000
	Topic: clickhouse_events_proto	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
Topic: clickhouse_person_unique_id	PartitionCount: 1	ReplicationFactor: 1	Configs: flush.ms=1000,segment.bytes=1073741824,flush.messages=10000,max.message.bytes=1000012,retention.bytes=15000000000
	Topic: clickhouse_person_unique_id	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
Topic: clickhouse_person	PartitionCount: 1	ReplicationFactor: 1	Configs: flush.ms=1000,segment.bytes=1073741824,flush.messages=10000,max.message.bytes=1000012,retention.bytes=15000000000
	Topic: clickhouse_person	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
Topic: plugin_log_entries	PartitionCount: 1	ReplicationFactor: 1	Configs: flush.ms=1000,segment.bytes=1073741824,flush.messages=10000,max.message.bytes=1000012,retention.bytes=15000000000
	Topic: plugin_log_entries	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
Topic: clickhouse_session_recording_events	PartitionCount: 1	ReplicationFactor: 1	Configs: flush.ms=1000,segment.bytes=1073741824,flush.messages=10000,max.message.bytes=1000012,retention.bytes=15000000000
	Topic: clickhouse_session_recording_events	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
Topic: __consumer_offsets	PartitionCount: 50	ReplicationFactor: 1	Configs: compression.type=producer,cleanup.policy=compact,flush.ms=1000,segment.bytes=104857600,flush.messages=10000,max.message.bytes=1000012,retention.bytes=15000000000
	Topic: __consumer_offsets	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
	Topic: __consumer_offsets	Partition: 1	Leader: 0	Replicas: 0	Isr: 0
...
	Topic: __consumer_offsets	Partition: 48	Leader: 0	Replicas: 0	Isr: 0
	Topic: __consumer_offsets	Partition: 49	Leader: 0	Replicas: 0	Isr: 0
Topic: events_dead_letter_queue	PartitionCount: 1	ReplicationFactor: 1	Configs: flush.ms=1000,segment.bytes=1073741824,flush.messages=10000,max.message.bytes=1000012,retention.bytes=15000000000
	Topic: events_dead_letter_queue	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
Topic: events_plugin_ingestion	PartitionCount: 1	ReplicationFactor: 1	Configs: flush.ms=1000,segment.bytes=1073741824,flush.messages=10000,max.message.bytes=1000012,retention.bytes=15000000000
	Topic: events_plugin_ingestion	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
```
Note how `events_plugin_ingestion	PartitionCount: 1` also that we might want to have different partition counts for different tables.

fuziontech:
`(Size of kafka disk * 0.7)/(sum over all T (P_T))` ≈ `(Size of kafka disk * 0.7)/(numPartitions * 2)` if you only have two large topics which was my point in the comment.

charts/posthog/values.yaml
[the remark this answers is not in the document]
fuziontech: (Size of kafka disk * 0.7)/(numPartitions * 2)
(x * 0.7)/(11 * 2) = 15
0.7x/22=15
0.7x = 22*15 = 330
x = 330 / 0.7 = about 470
Too big.

Target Kafka size of ~80gb means `logRetentionBytes` should be ~ `_2_540_000_000`
(80 * 0.7)/(11 * 2) = x = 2.54

fuziontech, line 452:
```suggestion
  logRetentionBytes: _2_540_000_000
```

fuziontech (approved):
Updated to new values for kafka PVC and `logRetentionBytes` as a compromise and increased durability of logs.

charts/posthog/values.yaml
```diff
@@ -446,12 +446,14 @@ kafka:
     # -- Enable persistence using PVC
     enabled: true
     # -- PVC Storage Request for kafka volume
-    size: 20Gi
+    size: 80Gi
```
tiina303, line 449: given our ClickHouse default of 20Gi (and other defaults) isn't this a bit high?

charts/posthog/values.yaml
```diff
@@ -446,12 +446,14 @@ kafka:
...
   # -- A size-based retention policy for logs
   # -- Should be less than kafka.persistence.size, minimum 1GB
-  logRetentionBytes: _15_000_000_000
+  logRetentionBytes: _2_333_333_333
```
tiina303, line 452: Add a comment explaining to users how they should calculate it if they change the size or partitions

charts/posthog/values.yaml
```diff
@@ -446,12 +446,14 @@ kafka:
     # -- Enable persistence using PVC
     enabled: true
     # -- PVC Storage Request for kafka volume
```
tiina303, line 448: Add a comment that they likely want to change `logRetentionBytes` too.

charts/posthog/values.yaml
```diff
@@ -446,12 +446,14 @@ kafka:
...
-    size: 20Gi
+    size: 80Gi
   # -- A size-based retention policy for logs
   # -- Should be less than kafka.persistence.size, minimum 1GB
-  logRetentionBytes: _15_000_000_000
+  logRetentionBytes: _2_333_333_333
   # -- The minimum age of a log file to be eligible for deletion due to age
   logRetentionHours: 24
+  # -- The default number of partitions per topic	
+  numPartitions: 12
```
guidoiaquinti, line 456: Can we make this a bigger number? Like 128?

tiina303:
Related docs page to update afterwards: https://posthog.com/docs/self-host/deploy/troubleshooting#why-did-we-run-into-this-problem-and-how-to-avoid-it-in-the-future

charts/posthog/values.yaml
yakkomajuri, line 456:
```suggestion
  numPartitions: 256
```

yakkomajuri:
Can we make the calculation dynamic? Decided for a really big partitions number (256 for now) because we can reap a lot of benefits here.

I have to make some changes to the main repo here.

Can someone help me with what's needed to get this over the line?

guidoiaquinti:
> Can we make the calculation dynamic? Decided for a really big partitions number (256 for now) because we can reap a lot of benefits here.

Unfortunately we can't. I'm 👍 to keep it static to a big number but 256 seems quite a lot. There are downsides of using too many partitions (more locks in ZK, more FD, etc...). 

What do you think about something like 64/128? 

> Can someone help me with what's needed to get this over the line?

I think we should verify a couple of things:

- retention settings (see above)
- if this change can be pushed at runtime with no issues
- make sure documentation is updated

fuziontech:
Thank you both to tending to this and +1 for getting it 🚢 ed

charts/posthog/values.yaml
```diff
@@ -446,12 +446,14 @@ kafka:
...
-    size: 20Gi
+    size: 80Gi
   # -- A size-based retention policy for logs
   # -- Should be less than kafka.persistence.size, minimum 1GB
-  logRetentionBytes: _15_000_000_000
+  logRetentionBytes: _2_333_333_333
   # -- The minimum age of a log file to be eligible for deletion due to age
   logRetentionHours: 24
+  # -- The default number of partitions per topic	
+  numPartitions: 256
```
yakkomajuri, line 456:
```suggestion
  numPartitions: 128
```

yakkomajuri:
Updated to 128. I'd love to enlist Team Infra's help here. I think we should really get this shipped and would love help to test and ship

@hazzadous @guidoiaquinti

guidoiaquinti:
You have a merge conflict to solve.

Testing: I think we should simply figure out if this change is disruptive or not and if yes address it accordingly (e.g. by adding a release note). First step is to probably spin up a stack from `main` and then do an update from this branch and see if everything works as expected (volumes are resized, retention policy gets modified, partitions are created, Kafka still works, plugin-server as well, etc..)

charts/posthog/values.yaml
```diff
@@ -481,7 +481,14 @@ kafka:
...
     # -- PVC Storage Request for Kafka data volume.
-    size: 20Gi
+    size: 80Gi
+  # -- A size-based retention policy for logs
+  # -- Should be less than kafka.persistence.size, minimum 1GB
+  logRetentionBytes: _2_333_333_333
+  # -- The minimum age of a log file to be eligible for deletion due to age
+  logRetentionHours: 24
+  # -- The default number of partitions per topic	
+  numPartitions: 128
```
macobo, line 491: This is way too many for anyone except people processing billions of events. We should be tuning these defaults for the average-sized companies.

charts/posthog/values.yaml
```diff
@@ -481,7 +481,14 @@ kafka:
     # - Enable data persistence using PVC.
     enabled: true
     # -- PVC Storage Request for Kafka data volume.
-    size: 20Gi
+    size: 80Gi
+  # -- A size-based retention policy for logs
+  # -- Should be less than kafka.persistence.size, minimum 1GB
+  logRetentionBytes: _2_333_333_333
```
macobo, line 487: This will cause out of disks since 128 * 2Gi is more than 80Gi

charts/posthog/values.yaml
```diff
@@ -481,7 +481,14 @@ kafka:
...
+  # -- The minimum age of a log file to be eligible for deletion due to age
+  logRetentionHours: 24
```
macobo, line 489: Problem to be solved here is ensuring that users don't run out of disk. I think we aren't really sure whether this is max(logRetentionBytes, logRetentionHours) or min.

charts/posthog/values.yaml
macobo, line 491: Also need to make sure that upgrading here is smooth for anyone coasting on mostly default values.

PR #244, opened 2022-01-11
dev(codespaces): add basic code spaces setup

hazzadous:
This just uses the github codespaces "universal" image, which has things
like kubectl, helm etc. and then adds in kind and creates a cluster.

I'm using the universal image with the hope that this will be caches on
the codespaces nodes :fingerscrossed:

## Description

## Type of change

## How has this been tested?

## Checklist

### DIFF
--- a/.devcontainer/setup_container.sh
+++ b/.devcontainer/setup_container.sh
@@ -1,0 +1,45 @@
+#/usr/bin/env bash
+
+set -e -o pipefail
+
+# Setup a cluster with kind + nginx ingress
+
+# Install kind so we can create a cluster as per
+# https://kind.sigs.k8s.io/docs/user/quick-start/#installing-from-release-binaries
+curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.11.1/kind-linux-amd64
+chmod +x ./kind
+mkdir -p ~/.local/bin/
+mv ./kind ~/.local/bin/kind
+
+# We use a cluster config that adds ingress-ready=true to each node as per
+# https://kind.sigs.k8s.io/docs/user/ingress/#create-cluster
+# 
+# This should mean you can go to port http://localhost:8000 or
+# https://localhost:4430 to view the running posthog app once deployed to the
+# cluster, and once VSCode has forwarded the ports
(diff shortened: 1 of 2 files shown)


guidoiaquinti (approved):
Non-blocking: can we add some info in the `REDAME.md` on how to use it?

.devcontainer/setup_container.sh
```diff
@@ -0,0 +1,45 @@
...
+    protocol: TCP
+  - containerPort: 443
+    hostPort: 4430
+    protocol: TCP
+EOF
+
+# Then provision the nginx controller, e.g.
+# https://kind.sigs.k8s.io/docs/user/ingress/#ingress-nginx 
+# NOTE: I've pinned the version here
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/d8c9a6c238f714587da4d2ac2dcd0d3d39419ccf/deploy/static/provider/kind/deploy.yaml
```
guidoiaquinti, line 43: Do we always need this?

.devcontainer/setup_container.sh
```diff
@@ -0,0 +1,45 @@
...
+
+echo "printf 'Hello 🦔! To install PostHog into k8s run this:\n\n "helm upgrade --install posthog charts/posthog --set cloud=private"\n'" >> ~/.zshrc
```
guidoiaquinti, line 45: This will not work outside `zsh`

.devcontainer/setup_container.sh
hazzadous: possibly not 🤔

.devcontainer/setup_container.sh
hazzadous: The image user is setup for zsh, I can change if you think useful?

hazzadous:
> Non-blocking: can we add some info in the `REDAME.md` on how to use it?

Sure!

.devcontainer/setup_container.sh
guidoiaquinti: Sorry, in that case make sense

PR #246, opened 2022-01-12
Remove global chart envs

tiina303:
## Description

Users can now add additional global envs into their values.yaml like this
```
env:
  - name: ADDITIONAL_ENV
    value: "false"
```
Instead of
```
env:
  - name: ASYNC_EVENT_PROPERTY_USAGE
    value: "true"
  - name: EVENT_PROPERTY_USAGE_INTERVAL_SECONDS
    value: "86400"
  - name: ADDITIONAL_ENV
    value: "false"
```

Depends on the PostHog version in use having this change https://github.com/PostHog/posthog/pull/7709
Fortunately that's the case for another PR already with a similar requirement: https://github.com/PostHog/charts-clickhouse/pull/179#issuecomment-1009872485

Opening as draft to make sure we don't land it too soon. Also made it a patch release with the assumption that we land it after 179 (which will be the major with PostHog version requirement).

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How has this been tested?

Will run CI e2e tests before landing.


guidoiaquinti:
I'm not sure I'm following the purpose of this PR. Is the plan to put all the defaults in `posthog/posthog` and remove the overrides here?

tiina303:
> I'm not sure I'm following the purpose of this PR. Is the plan to put all the defaults in posthog/posthog and remove the overrides here?

yes, I did that in the PR linked. The purpose is (1) making it easier/clearer for users how they can add overrides to envs. Sure event without this PR you could do 
```
env:
  - name: ADDITIONAL_ENV
    value: "false"
```
As the defaults are right in posthog/posthog, but I wouldn't assume I could do that looking at the `values.yaml`. So reducing confusion. Additionally (2) if we ever want to change any defaults for PostHog that should be in the code there as we can't easily tie chart version to PostHog version (I mean there are no guarantees), so this removes a bad pattern.

charts/posthog/values.yaml
```diff
@@ -31,11 +31,7 @@ clickhouseOperator:
   serviceType: NodePort
 
 # -- Env vars to throw into every deployment (web, worker, and plugin server)
-env:
-  - name: ASYNC_EVENT_PROPERTY_USAGE
-    value: "true"
-  - name: EVENT_PROPERTY_USAGE_INTERVAL_SECONDS
-    value: "86400"
+env: []
```
fuziontech, line 34: I would add a comment here to show how to add env vars if they want to. The [] is a bit confusing.

Just an example

```yaml
# env:
#   - name: FOO
#     value: bar
```

fuziontech (approved):
:shipit: but just add the comment (mentioned in the comment)

I guess that's a meta comment or higher order comment? :P

charts/posthog/values.yaml
tiina303, line 34:
```suggestion
env: []
# env: 
#   - name: FOO 
#     value: bar
```

tiina303: Note that we do have an example later in the file now, but added since I though about it earlier too

tiina303:
Ran the e2e test just in case too [AWS](https://github.com/PostHog/charts-clickhouse/actions/runs/1922189625), [DO](https://github.com/PostHog/charts-clickhouse/actions/runs/1922192879), [GCP](https://github.com/PostHog/charts-clickhouse/actions/runs/1922191622)

PR #248, opened 2022-01-18
Redis - refactor ENV variables injection & secret management

guidoiaquinti:
## Description
This PR  refactors ENV variables injection & secret management for the Redis service.

The main changes are:

1. redefined/added helpers variables related to Redis
2. centralised ENV variables injection into a single snippet that can be included in different resource definitions
3. introduced a new k8s`Secret` resource that will be used to store the password provided via `values.yaml`
4. `redis.host` `redis.port` `redis.password` are now `externalRedis.host` `externalRedis.port` `externalRedis.password`
5. added support for existing secrets via the input `externalRedis.existingSecret` and `externalRedis.existingSecretPasswordKey`
6. added integration tests via Kubetest to all the scenarios above

Here's the PR with the upgrade notes https://github.com/PostHog/posthog.com/pull/2804

## Type of change

## How has this been tested?
CI is passing ✅ 

## Checklist
- [X] I have commented my code, particularly in hard-to-understand areas


ci/values/k3s.yaml
```diff
@@ -27,3 +23,6 @@ redis:
   master:
     persistence:
       size: 1Gi
+zookeeper:
+  persistence:
+    size: 1Gi
```
macobo, line 28: Is this a breaking change for users with defaults? Will this cause trouble for them?

macobo (commented):
Took a look from my side - this looks solid, especially the test coverage. The only concern I had was the zookeeper defaults change.

charts/posthog/templates/_helpers.tpl
```diff
@@ -171,66 +162,97 @@ Set pgbouncer URL
...
-
 
 {{/*
-Set redis port
+Return the Redis secret key
 */}}
-{{- define "posthog.redis.port" -}}
-{{- if .Values.redis.enabled -}}
-    6379
+{{- define "posthog.redis.secretPasswordKey" -}}
```
macobo, line 233: We do a clever trick here - checking for keys existing. Should we do the same trick to validate that users aren't passing in old, removed keys (redis.host, etc)?

ci/values/k3s.yaml
guidoiaquinti: > Is this a breaking change for users with defaults? Will this cause trouble for them?

This is not, it is only an override for the `k3s` test

> Also this seems undocumented in the main values.yaml file.

Correct, we do not expose all the variables from subcharts (we probably should)

charts/posthog/templates/_helpers.tpl
guidoiaquinti: We had a similar concern for the PR that deprecated custom `StorageClass`es and like there I'm inclined to don't add any additional check and simply provide detailed information in the release notes.

ci/values/k3s.yaml
macobo: Ah, didn't notice the CI dir here.

charts/posthog/templates/_helpers.tpl
macobo: I think it's worth revisting this: people will skip reading release notes or miss things they need to update, which in turn will translate to a broken experience and more support burden for us.

We could create a single block somewhere which for every removed value gives some guidance that this is now removed. This shouldn't be too much of a maintanence burden but would get us most of the way there in terms of UX.

guidoiaquinti: > people will skip reading release notes or miss things they need to update, which in turn will translate to a broken experience and more support burden for us.

I agree but in that case Helm should automatically rollback to the previous version. Also the new pods will not pass the readiness check so also in that case no downtime / no broken experience is expected.

macobo: > I agree but in that case Helm should automatically rollback to the previous version

Can we guarantee that happens every time though? In any case, PRs over issues - if you are willing to delay the major version release a bit, will have a PR for you soon :)

macobo: https://github.com/PostHog/charts-clickhouse/pull/261

ci/kubetest/test_redis_external.py
```diff
@@ -0,0 +1,50 @@
...
+@pytest.fixture
+def setup(kube):
+    cleanup_k8s()
+    create_namespace_if_not_exists()
+    install_custom_resources("./custom_k8s_resources/redis_external.yaml")
+    helm_install(HELM_INSTALL_CMD)
+    wait_for_pods_to_be_ready(kube)
+
+
+def test_helm_install(setup, kube):
```
hazzadous, line 39: non-blocking

Not in love with making tests dependent. Maybe we could go with making the setup fixture scope="module" instead?

ci/kubetest/test_redis_external_with_existing_secret.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_posthog_healthy(kube):
+    is_posthog_healthy(kube)
+
+
+def test_redis_secret(kube):
+    secrets = kube.get_secrets(namespace="posthog", fields={"type": "Opaque"})
+
+    default_redis_secret_name = "posthog-posthog-redis-external"
```
hazzadous, line 50: non-blocking

Seems like this might be a bit brittle e.g. we change he default name, the test doesn't fail. Could we instead validate that the existing secret is used? Fine if not.

ci/kubetest/test_redis_external.py
guidoiaquinti: https://github.com/PostHog/charts-clickhouse/issues/265

hazzadous (approved):
lgtm

ci/kubetest/test_redis_external_with_existing_secret.py
guidoiaquinti: If you agree I'm going to fix this separately so we can land this in the meantime.

PR #254, opened 2022-01-21
Fix web-hpa config

guidoiaquinti:
## Description
With https://github.com/PostHog/charts-clickhouse/pull/179 we've removed by mistake the `name` metadata field in the web hpa template. This can causes errors on upgrade if HPA is enabled (thank you @legion2 for [the report](https://github.com/PostHog/charts-clickhouse/pull/179#discussion_r789451655)).

## Type of change
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## How has this been tested?
CI should be passing

## Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #256, opened 2022-01-21
CI - add test for HPA enabled

guidoiaquinti:
## Description
Basic test (we only verify `helm upgrade --install`) for our HPA setup. 

We are adding it to catch future regressions like https://github.com/PostHog/charts-clickhouse/pull/254 

## Type of change

## How has this been tested?
CI ✅ 

## Checklist

### DIFF
--- a/ci/kubetest/test_clickhouse_persistence_disabled.py
+++ b/ci/kubetest/test_clickhouse_persistence_disabled.py
@@ -15,7 +15,7 @@ helm upgrade \
 """
 
 
-def test_helm_install(setup, kube):
+def test_helm_install(kube):
     cleanup_k8s()
     helm_install(HELM_INSTALL_CMD)
     wait_for_pods_to_be_ready(kube)
--- a/ci/kubetest/test_clickhouse_persistence_disabled.py
+++ b/ci/kubetest/test_clickhouse_persistence_disabled.py
@@ -15,15 +15,10 @@ helm upgrade \
 """
 
 
-@pytest.fixture
-def setup(kube):
+def test_helm_install(setup, kube):
     cleanup_k8s()
     helm_install(HELM_INSTALL_CMD)
     wait_for_pods_to_be_ready(kube)
-
-
-def test_helm_install(setup, kube):
-    pass
 
 
 def test_volume_claim(kube):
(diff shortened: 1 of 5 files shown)


ci/kubetest/test_clickhouse_persistence_disabled.py
```diff
@@ -22,7 +22,11 @@ def setup(kube):
     wait_for_pods_to_be_ready(kube)
 
 
-def test_volume_claim(setup, kube):
+def test_helm_install(setup, kube):
```
hazzadous, line 25: This looks a bit funny with the actual bit being tested in a fixture, could we instead run the commands within the test function?

I'm becoming more of the opinion that pytest fixture are cute, but not obvious

ci/kubetest/test_clickhouse_persistence_disabled.py
```diff
@@ -15,14 +15,13 @@
...
-@pytest.fixture
-def setup(kube):
+def test_helm_install(setup, kube):
     cleanup_k8s()
     helm_install(HELM_INSTALL_CMD)
     wait_for_pods_to_be_ready(kube)
 
 
-def test_volume_claim(setup, kube):
+def test_volume_claim(kube):
```
hazzadous, line 24: Does this mean we rely on `test_helm_install` having run prior to this one?

ci/kubetest/test_clickhouse_persistence_disabled.py
```diff
@@ -15,14 +15,13 @@
 """
 
 
-@pytest.fixture
-def setup(kube):
+def test_helm_install(setup, kube):
```
hazzadous, line 18: Where is the `setup` fixture defined?

ci/kubetest/test_clickhouse_persistence_disabled.py
guidoiaquinti: Yes

hazzadous: I think we should just drop the non-hpa test changes and get this in, is that possible?

I think test interdependence is something we should avoid, as will make it hard to e.g. like you said introduce parallelism among other things

hazzadous (approved):

PR #257, opened 2022-01-21
Fix template defaults for ClickHouse

guidoiaquinti:
## Description
With https://github.com/PostHog/charts-clickhouse/pull/224 we moved the `ClickHouseInstallation` defaults to be cluster specific but we introduced an unexpected regression (due to the `clusterServiceTemplate` -> `serviceTemplate` difference) that we are now fixing. 

Instead of using the custom settings we specify in `service-template` we were instead using the defaults.

## Type of change

## How has this been tested?
CI is ✅ 

## Checklist


tiina303 (approved):
lgtm

PR #261, opened 2022-01-24
Proposal: Raise errors if using now-unsupported values

macobo:
## Description

This PR makes it so we raise errors when values we no longer support are used during upgrades. This helps catch easier bugs during updates.

Note the target branch is set to https://github.com/PostHog/charts-clickhouse/pull/248 as the PR arose from a discussion under it.

<details><summary>Click here to see example errors</summary>

```
⟩ helm upgrade --install -f ci/values/k3s.yaml posthog charts/posthog --timeout 30m --set clickhouseOperator.enabled=true                                                                                  
Error: UPGRADE FAILED: template: posthog/templates/web-deployment.yaml:1:4: executing "posthog/templates/web-deployment.yaml" at <include "snippet.removed-vars-errors" .>: error calling include: template: posthog/templates/_snippet-removed-vars-errors.tpl:6:8: executing "snippet.removed-vars-errors" at <required (printf (include "snipper.removed-vars-errors-template" .) "clickhouseOperator values are no longer valid." "https://posthog.com/docs/self-host/deploy/upgrade-notes#upgrading-from-9xx") nil>: error calling required: HELM_ERR_START


==== BREAKING CHANGE ====

clickhouseOperator values are no longer valid.

Read more on how to update your values.yaml:
https://posthog.com/docs/self-host/deploy/upgrade-notes#upgrading-from-9xx

=========================

HELM_ERR_END

```

</details>

Question(s) to reviewers:
- Should we add any other variables to this list?
- Should I add any other tests to this? If so, would love a bit of guidance on _what_ kind of test would be appropriate or other guidance.

~Also note we could also improve the formatting - the error message tends to get lost among everything else, but we could strategically insert new lines and whatnot. Suggestions for how to bannerize this are welcome!~ Added a solution.

## Type of change

## How has this been tested?

## Checklist


charts/posthog/templates/_snippet-removed-vars-errors.tpl
```diff
@@ -0,0 +1,24 @@
+{{/* Checks whether removed variables are in use. */}}
```
guidoiaquinti, line 1: Should we simply add this to `_helpers.tpl` ?

macobo: I think it's more likely it will get lost in the noise and never updated if it was in that file.

guidoiaquinti:
Meta question on this but I'm not super sure we should intercept and error on inputs that are != posthog related. 

I know that in the past we didn't do any separation at the values file level (e.g. `redis.*` variables are only for the `redis` chart, `kafka.*` are only for the `kafka` chart,...) but as we are now cleaning up stuff and moving into that direction, I'm not sure if this PR will help (as we intercept and prevent overrides in a "namespace" that is not posthog)

guidoiaquinti (approved):
CI test failures are unrelated

PR #267, opened 2022-01-25
clickhouse-operator config map: fix `listen_host`

macobo:
## Description

I couldn't get the chart working locally with `kind`. Digging in, the
issue seemed to be that clickhouse failed to bind to 0.0.0.0.

This change makes it so we don't try to double-bind - :: already covers
0.0.0.0. As to why this only affected my setup - no clue, but the bug
seems pretty obvious.

Related docs link: https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings/#server_configuration_parameters-listen_host

## Type of change

## How has this been tested?

I removed the lines, uninstalled and re-ran helm install. It now worked.

Note:
- this does _not_ restart the clickhouse pod
- configmap does not seem to get updated automatically when it changes. In essence this change only affects new installs
  - When I added a line to the file and ran `kubectl exec --stdin --tty chi-posthog-posthog-0-0-0 -n posthog -- cat /etc/clickhouse-server/config.d/01-clickhouse-01-listen.xml` the contents were the same. 

## Checklist

cc @guidoiaquinti - would love some guidance on what level of testing do you think is appropriate here.

### DIFF
--- a/charts/posthog/templates/clickhouse-operator/configmap.yaml
+++ b/charts/posthog/templates/clickhouse-operator/configmap.yaml
@@ -215,8 +215,6 @@ data:
     <yandex>
         <!-- Listen wildcard address to allow accepting connections from other containers and host network. -->
         <listen_host>::</listen_host>
-        <listen_host>0.0.0.0</listen_host>
-        <listen_try>1</listen_try>
     </yandex>
 
   01-clickhouse-02-logger.xml: |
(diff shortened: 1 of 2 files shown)


guidoiaquinti:
> would love some guidance on what level of testing do you think is appropriate here.

I think we could add a check in the current `kubetest` suite to verify if `curl -g -6 $ENDPOINT` and `curl -g -4 $ENDPOINT` works (to verify both ipv4 and ipv6 binding)

guidoiaquinti (approved):

Issue #249, opened 2022-01-18
Make clickhouse settings configurable


macobo:
### Proposed change

Currently, clickhouse settings are hard-coded in the chart. They should be configurable by us and for chart users - e.g. to increase default memory settings and others

### Alternative options

???

## Additional context

We set these settings currently on cloud: https://github.com/PostHog/posthog-cloud-infra/blob/main/ansible/config/clickhouse-users.xml

Open questions: How will this affect existing installations if these values change?

PR #268, opened 2022-01-25
Allow passing custom clickhouse settings

macobo:
## Description

This PR fixes https://github.com/PostHog/charts-clickhouse/issues/249 by making clickhouse settings configurable.

Companion documentation PR: https://github.com/PostHog/posthog.com/pull/2833

## Type of change

## How has this been tested?

See tests. I also tested manually. One edge case I ran into was when passing invalid settings, clickhouse doesn't automatically update. If then killed, the pod won't start up again until the issue is fixed.

## Checklist

### DIFF
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -12,7 +12,7 @@ spec:
       admin/quota: default
 
     profiles:
-      {{- merge dict .Values.clickhouse.defaultProfiles .Values.clickhouse.profiles | toYaml | nindent 6 }}
+      {{- merge dict .Values.clickhouse.profiles .Values.clickhouse.defaultProfiles | toYaml | nindent 6 }}
 
     clusters:
       - name: {{ .Values.clickhouse.database | quote }}
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -13,6 +13,9 @@ spec:
 
     profiles:
       default/allow_experimental_window_functions: "1"
+      {{- if .Values.clickhouse.profiles }}
+      {{- toYaml .Values.clickhouse.profiles | nindent 6 }}
+      {{- end }}
 
     clusters:
       - name: {{ .Values.clickhouse.database | quote }}
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -12,10 +12,7 @@ spec:
       admin/quota: default
 
     profiles:
-      default/allow_experimental_window_functions: "1"
-      {{- if .Values.clickhouse.profiles }}
-      {{- toYaml .Values.clickhouse.profiles | nindent 6 }}
-      {{- end }}
+      {{- merge dict .Values.clickhouse.defaultProfiles .Values.clickhouse.profiles | toYaml | nindent 6 }}
 
     clusters:
       - name: {{ .Values.clickhouse.database | quote }}
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -8,7 +8,7 @@ image:
   sha:
   # -- Posthog image tag, e.g. release-1.31.1
   tag:
-  # -- Default image or tag, e.g. `:release-1.31.1` Do not overwrite!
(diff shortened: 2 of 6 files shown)


charts/posthog/templates/clickhouse_instance.yaml
```diff
@@ -13,6 +13,9 @@ spec:
 
     profiles:
       default/allow_experimental_window_functions: "1"
```
guidoiaquinti, line 15: As discussed on Slack, let's put this as a common default in `values.yaml` and then merge that map/dict with `.Values.clickhouse.profiles`

ci/kubetest/test_clickhouse_custom_settings_profile.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.fixture
+def setup(kube):
+    cleanup_k8s()
+    install_chart(VALUES_YAML)
+    wait_for_pods_to_be_ready(kube)
+
+
+def test_custom_clickhouse_settings_profile(setup, kube):
+    command = f"clickhouse-client -q \"SELECT name, value FROM system.settings WHERE name IN ('max_execution_time', 'max_memory_usage') ORDER BY name FORMAT JSON\""
+    result = kubectl_exec("chi-posthog-posthog-0-0-0", command)
```
guidoiaquinti, line 54: Nit: not sure about it but `kubetest` might expose a method on the container object for this already

ci/kubetest/utils.py
```diff
@@ -92,18 +119,28 @@ def test_if_posthog_deployments_are_healthy(kube):
...
+
+
+def _exec_subprocess(cmd):
+    cmd_run = subprocess.run(cmd, shell=True, capture_output=True, text=True)
     cmd_return_code = cmd_run.returncode
     if cmd_return_code:
-        pytest.fail("❌ Error while running '{}'. Return code: {}".format(cmd, cmd_return_code))
-    log.debug("✅ Done!")
+        pytest.fail(
+            f"""
```
guidoiaquinti, line 138: maybe it's GH but doesn't seems to be indented properly

macobo: indented by black, so should be OK

ci/kubetest/test_clickhouse_custom_settings_profile.py
macobo: https://kubetest.readthedocs.io/en/latest/reference.html#pod Doesn't seem like it. However I could probably use the HTTP API for this + http_proxy methods. Will experiment!

macobo: This almost works - ran into a hitch where authentication always fails with this clickhouse-side error:

```
2022.01.25 15:09:23.030568 [ 60 ] {} <Warning> AddressPatterns: Failed to check if the allowed client hosts contain address ::ffff:10.244.0.1. DB::Exception: Cannot getnameinfo(::ffff:10.244.0.1): Name or service not known, code = 198
```

Will need to dig deeper tomorrow.

charts/posthog/templates/clickhouse_instance.yaml
```diff
@@ -12,7 +12,7 @@ spec:
       admin/quota: default
 
     profiles:
-      default/allow_experimental_window_functions: "1"
+      {{- merge dict .Values.clickhouse.profiles .Values.clickhouse.defaultProfiles | toYaml | nindent 6 }}
```
guidoiaquinti, line 15: not sure about the ordering (and if something changes) but should we do default first and then overrides?

charts/posthog/values.yaml
```diff
@@ -8,7 +8,7 @@ image:
   sha:
   # -- Posthog image tag, e.g. release-1.31.1
   tag:
-  # -- Default image or tag, e.g. :release-1.31.1
```
guidoiaquinti, line 11: Why?

macobo: Should override one of the two above settings instead. Could make that clear in the comment though

ci/kubetest/test_clickhouse_custom_settings_profile.py
macobo: Dug a bit more - I don't think it's worth investing more here, seems like a deeper clickhouse-operator related bug than worth tackling here.

guidoiaquinti (approved):

PR #271, opened 2022-01-25
Kafka - refactor ENV variables & config

guidoiaquinti:
## Description
This PR refactor all the Kafka ENV variables & config. Main changes are:

1. the built-in Kafka service type default is now `ClusterIP` from the previous `NodePort`. I've also removed the custom Service we were deploying as we can simply relying on the one provided by `bitnami/kafka`

2. the values `kafka.url`, `kafka.host`, `kafka.port` have been consolidated into `externalKafka.brokers` (example: `externalKafka.brokers: "kafka-broker1:9094,kafka-broker2:9094,kafka-broker3:9094"`)

Release notes are available at: https://github.com/PostHog/posthog.com/pull/2826

## Type of change

## How has this been tested?
CI is ✅ 

## Checklist


charts/posthog/values.yaml
```diff
@@ -1,6 +1,12 @@
-# Default values for posthog.
-# This is a YAML-formatted file.
-# Declare variables to be passed into your templates.
+## @section Helm parameters
```
macobo, line 1: This doesn't work with helm-docs last I checked. What is this formatting?

charts/posthog/templates/kafka-service.yaml
```diff
@@ -1,21 +0,0 @@
-{{- if .Values.kafka.service.enabled }}
```
guidoiaquinti, line 1: Not sure why this was added in the first place but we have a built-in service available with the upstream chart.

charts/posthog/values.yaml
guidoiaquinti: I was thinking to slowly move to https://github.com/bitnami-labs/readme-generator-for-helm as it looks a bit more complete in my opinion (happy to chat if you think we shouldn't do it!)

ci/kubetest/test_clickhouse_persistence_enabled_existing_claim.py
```diff
@@ -22,20 +28,11 @@
...
-        pytest.fail("❌ Error while running '{}'. Return code: {}".format(cmd, cmd_return_code))
-    log.debug("✅ Done!")
-
-
 @pytest.fixture
 def setup(kube):
     cleanup_k8s()
-    create_custom_pvc()
+    create_namespace_if_not_exists()
+    install_custom_resources("./custom_k8s_resources/clickhouse_existing_claim.yaml")
```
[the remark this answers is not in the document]
macobo: Also a potential merge conflict with https://github.com/PostHog/charts-clickhouse/pull/267 - ran into the same bug.

ci/kubetest/test_kafka_service_nodeport.py
```diff
@@ -0,0 +1,31 @@
...
+    wait_for_pods_to_be_ready(kube)
+
+
+def test_helm_install(setup, kube):
+    pass
+
+
+def test_kafka_service(kube):
+    service = get_kafka_service_spec(kube)
+    assert service.type == "NodePort", "Kafka service type is {}".format(service.type)
```
macobo, line 31: Does this give us enough information that we can actually connect to the service from the relevant pods?

I assume not right?

I think it would be worthwhile to smoke test this by:
- Spin up a web pod
- Test accessibility either via https://github.com/PostHog/posthog/pull/8258 or via some other method that actually connects to the kafka cluster.

ci/values/kubetest/test_kafka_service_nodeport.yaml
```diff
@@ -0,0 +1,34 @@
+cloud: local
+
+kafka:
+  enabled: true
+  service:
+    type: NodePort
```
macobo, line 6: Note: I don't think this test gives us anything that a unit test wouldn't give us.

Suggestion: Drop this particular test (but make sure it's covered by unit tests), test connectivity from within pods within the external kafka test.

charts/posthog/values.yaml
macobo: No strong opinions on tools to use and the tool looks good - only thing is that we'll break ALL_VALUES.md quite bad with this change. 

If you're sold on the tool, let's switch over ASAP, but in a separate PR.

ci/kubetest/test_kafka_service_nodeport.py
guidoiaquinti: 100% the PR is still a WIP so I was going to iterate further on this 😄 

As of now, this test doesn't make sense as kubetest as we are testing something that we should get as "granted" from the upstream chart. I also don't think we should have a unit test specifically for this for the same reason.

If we consider adding as well some additional test condition (like the one you proposed) I think it might make sense.

ci/values/kubetest/test_kafka_service_nodeport.yaml
guidoiaquinti: Ditto

charts/posthog/values.yaml
guidoiaquinti: Ack, reverted (added as well fixes for the Redis and ClickHouse sections I've edited few PRs ago)

ci/kubetest/test_kafka_service_nodeport.py
guidoiaquinti: Removed

ci/values/kubetest/test_kafka_service_nodeport.yaml
guidoiaquinti: Removed

ci/kubetest/test_kafka_external.py
```diff
@@ -0,0 +1,60 @@
...
+)
+
+VALUES_YAML = """
+cloud: local
+
+kafka:
+  enabled: false
+
+externalKafka:
+  brokers: "kafka0:9092"
```
hazzadous, line 19: Non-blocking

Maybe not relevant, but can we add a test that uses comma separated list here.

(I see we're doing some manipulation for the dependency check but that's only for internal, which I assume is always going to be a single clusterip? So that isn't an issue, but for e.g. MSK we'll have a list.)

ci/kubetest/test_kafka_external.py
```diff
@@ -0,0 +1,60 @@
...
+    install_chart(VALUES_YAML)
+    wait_for_pods_to_be_ready(kube)
+
+
+def test_helm_install(setup, kube):
+    pass
+
+
+def test_posthog_healthy(kube):
+    is_posthog_healthy(kube)
```
hazzadous, line 60: non-blocking

I wonder if we should also somehow check that if kafka isn't configured correctly then we aren't healthy. But perhaps that's just an assumption we need to make about the app interface.

hazzadous (approved):
Just a question about confidence of what we're checking. Non-blocking, looks good.

charts/posthog/templates/_kafka.tpl
```diff
@@ -0,0 +1,31 @@
...
+    {{- printf "%s" .Values.externalKafka.brokers -}}
+{{- end -}}
+{{- end -}}
+
+{{/* ENV used by PostHog deployments for connecting to Kafka */}}
+{{- define "snippet.kafka-env" }}
+- name: KAFKA_HOSTS
+  value: "{{ include "posthog.kafka.brokers" . }}"
+- name: KAFKA_URL
+  value: "kafka://{{ include "posthog.kafka.brokers" . }}"
```
hazzadous, line 28: The format posthog is expecting "kafka://broker0:123,kafka://broker1:123"

This may work assuming the app picks up the first correctly, but I think will not pick up the second: see https://github.com/PostHog/posthog/blob/2e6aae702d3e702188dad0e72e547ed6525ab638/posthog/settings/data_stores.py#L108

hazzadous (changes requested):
Just spotted and commented on KAFKA_URL format. 🤞 the test fails

guidoiaquinti:
> Just spotted and commented on KAFKA_URL format. 🤞 the test fails

Nice catch! Unfortunately the test didn't fail because I think the driver will use the first broker to bootstrap and fetch the others. This is something we need to fix but after discussing with @macobo we are inclined to merge this now **without** cutting a release so we can unlock few other workstreams currently depending on this change. We will then work to make the appropriate changes to fix the bug reported + cut a new release + write docs.

macobo (commented):
:+1:

PR #273, opened 2022-01-26
feat(static-assets): allow specifying djangos `STATIC_URL`

hazzadous:
To take advantage of e.g. a CDN we want to be able to specify an
external url from which clients should retrieve static assets.

## Description

## Type of change

## How has this been tested?

## Checklist

### DIFF
--- a/charts/posthog/templates/web-deployment.yaml
+++ b/charts/posthog/templates/web-deployment.yaml
@@ -69,6 +69,8 @@ spec:
         {{- include "snippet.redis-env" . | indent 8 }}
         - name: SITE_URL
           value: {{ template "posthog.site.url" . }}
+        - name: STATIC_URL
+          value: {{ default "/static/" .Values.web.staticUrl | quote }}
         - name: DEPLOYMENT
           value: {{ template "posthog.deploymentEnv" . }}
         - name: SENTRY_DSN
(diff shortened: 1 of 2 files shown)


charts/posthog/values.yaml
```diff
@@ -95,6 +95,8 @@ web:
     # requests:
     #   cpu: 300m
     #   memory: 300Mi
+  # -- Location from which application static assets (js, css, etc.) should be served
+  staticUrl: "/static/"
```
macobo, line 99: staticAssetsUrl?

charts/posthog/templates/web-deployment.yaml
```diff
@@ -69,6 +69,8 @@ spec:
         {{- include "snippet.redis-env" . | indent 8 }}
         - name: SITE_URL
           value: {{ template "posthog.site.url" . }}
+        - name: STATIC_URL
```
macobo, line 72: I'd also add this to events-deployment if only to keep the env vars in sync to avoid copy-pasta errors down the line :)

charts/posthog/values.yaml
hazzadous: Yeah I was mirroring the django env var, but staticAssetsUrl is a bit clearer out of django context.

charts/posthog/templates/web-deployment.yaml
hazzadous: Yup will check where else this might be relevant, I guess just anywhere we're running the django app

PR #282, opened 2022-01-31
Revert "clickhouse-operator config map: fix `listen_host`"

hazzadous:
This fix updated the clickhouse chart to not bind on `0.0.0.0`, but only the [ipv6 `0:0:0:0:0:0:0:0` unspecified address](https://www.rfc-editor.org/rfc/rfc3513#section-2.5.2), or `::` for short.

If we do not have ipv6 enabled we will have issues, as we will not be able to bind to any address. The resulting error from clickhouse is e.g.:

```
2022.01.31 14:36:07.584143 [ 7 ] {} <Error> Application: DB::Exception: Listen [::]:8123 failed: Poco::Exception. Code: 1000, e.code() = 0, e.displayText() = DNS error: EAI: Address family for hostname not supported (version 21.6.5.37 (official build))
```

We want to ensure that we also try the ipv4 0.0.0.0. listen_try 1 will ensure that we do not fail to load as long as we succeed in binding to at least one of the listen_hosts. This way we should work on either ipv4 or ipv6 disabled systems.

This is an initial revert just to verify that this indeed resolves this test [failure here for deploying to an aws eks test cluster](https://github.com/PostHog/charts-clickhouse/runs/5004910550?check_suite_focus=true).

I'll look into the initial issue this PR resolved as it would be good not to regress.

There is an option where we say we do not support ipv6 disabled clusters, but I suspect that will limit our compat. considerable.

Reverts PostHog/charts-clickhouse#267

### DIFF
--- a/charts/posthog/templates/clickhouse-operator/configmap.yaml
+++ b/charts/posthog/templates/clickhouse-operator/configmap.yaml
@@ -215,6 +215,8 @@ data:
     <yandex>
         <!-- Listen wildcard address to allow accepting connections from other containers and host network. -->
         <listen_host>::</listen_host>
+        <listen_host>0.0.0.0</listen_host>
+        <listen_try>1</listen_try>
     </yandex>
 
   01-clickhouse-02-logger.xml: |
--- a/charts/posthog/tests/clickhouse-operator/__snapshot__/configmap.yaml.snap
+++ b/charts/posthog/tests/clickhouse-operator/__snapshot__/configmap.yaml.snap
@@ -188,6 +188,8 @@ the manifest should match the snapshot when using default values:
         <yandex>
             <!-- Listen wildcard address to allow accepting connections from other containers and host network. -->
             <listen_host>::</listen_host>
+            <listen_host>0.0.0.0</listen_host>
+            <listen_try>1</listen_try>
         </yandex>
       01-clickhouse-02-logger.xml: |
         <yandex>


guidoiaquinti (approved):

macobo:
> listen_try 1 will ensure that we do not fail to load as long as we succeed in binding to at least one of the listen_hosts. **This way we should work on either ipv4 or ipv6 disabled systems.**

This isn't the case. The previous fix is needed to run the chart under my kind/linux setup as documented in https://github.com/PostHog/charts-clickhouse/pull/267. Interested in figuring out a proper fix.

@hazzadous how to repro the issue you were having? Do you have a test environment for example?

hazzadous:
> This isn't the case. The previous fix is needed to run the chart under my kind/linux setup as documented in #267. Interested in figuring out a proper fix.

Right, I'm not sure what is going on in my setup then. I'm also using kind/linux as setup as per [this script](https://github.com/PostHog/charts-clickhouse/blob/main/.devcontainer/setup_container.sh#L9) using the [ms universal linux docker image](https://github.com/PostHog/charts-clickhouse/blob/main/.devcontainer/devcontainer.json#L5), but there are lot's of layers of abstraction in my setup which would make it not a simple example. I'll try locally to see what I get. @guidoiaquinti did you have issues with the bind as well?

> @hazzadous how to repro the issue you were having? Do you have a test environment for example?

Where I noticed this was when deploying to the cluster generated by [this terraform](https://github.com/PostHog/posthog-cloud-infra/blob/main/terraform/environments/aws-accnt-staging/us-east-1/main.tf#L32), but it also appears to be failing on the aws test, as per the link in the PR description. Perhaps this was a flake though, I'll try running it again.

macobo:
That's still helpful - given my setup is the one that's currently failing, will investigate more on my end and perhaps reach out once I think I have a better fix! :)

Issue #280, opened 2022-01-31
Chart does not work without pgbouncer.enabled


macobo:
## Bug description

`pbouncer.enabled` exists, but is not respected by the chart.

E.g. the following env vars are always set by pgbouncer values:

```
        - name: POSTHOG_POSTGRES_HOST
          value: {{ template "posthog.pgbouncer.host" . }}
        - name: POSTHOG_POSTGRES_PORT
          value: {{ include "posthog.pgbouncer.port" . | quote }}
```

## Expected behavior

If pgbouncer is disabled, these env vars are populated by postgresql values.

## Actual behavior

Unavailable service is used

## Additional context

Related issues: 
- we expose `postgresql.existingSecret`, but `postgresql.password` is always used instead in some connection URLs.
- USING_PGBOUNCER is currently always hardcoded.

Legion2:
I would like to disable pgbouncer, because we don't need it

PR #285, opened 2022-01-31
PostgreSQL support improvements

macobo:
## Description

This PR refactors chart postgresql support. The following changes were made:
- (breaking) `externalPostgresql` has been added, `postgresql.postgresqlHost` and `postgresql.postgresqlPort` have been moved under it
- (breaking) `postgresql.postgresqlUsername` was removed. Posthog currently requires postgres admin access to install extensions and the subchart only gives admin permissions to the default `postgres` user.
- (breakingish) `postgresql.existingSecretKey` has been removed. Note this never worked and isn't supported by the subchart.
- (fix) `postgresql.existingSecret` did previously not work
- (testing) There's a new postgresql testing suite checking connectivity under various configurations

There are other known issues relating to postgres still remaining though. I've documented these in https://github.com/PostHog/charts-clickhouse/issues/280

I didn't manage to dig into RDS/TLS support as part of this PR - will need to be a separate PR (or just merging james' PR)

Documentation: https://github.com/PostHog/posthog.com/pull/2872

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How has this been tested?

See kubetest.

## Checklist


charts/posthog/templates/_postgresql.tpl
```diff
@@ -99,29 +100,50 @@ Set postgres port
...
+{{- if and .Values.postgresql.enabled (not .Values.postgresql.existingSecret) -}}
+{{- .Values.postgresql.postgresqlPassword -}}
+{{- else if and (not .Values.postgresql.enabled) (not .Values.externalPostgresql.existingSecret) -}}
+{{- .Values.externalPostgresql.postgresqlPassword -}}
+{{- end -}}
 {{- end -}}
 
 {{/*
-Set postgres URL
+Set postgres secret created in secrets.yaml
```
guidoiaquinti, line 143: I think we should not set this anymore and simply provide the password via env variables (like we do for redis)

charts/posthog/values.yaml
```diff
@@ -334,24 +334,32 @@ postgresql:
...
     enabled: true
     # -- PVC Storage Request for PostgreSQL volume
     size: 10Gi
-  # -- Host postgres is accessible from. Only set when internal PG is disabled
+  # -- Secret containing desired postgresql password. Secret must be behind key `postgresql-password`.  When defined the `postgresqlPassword` field is ignored
+  existingSecret:
+
+externalPostgresql:
+  # -- Host postgresql is accessible from
   postgresqlHost:
```
guidoiaquinti, line 349: nit: I would use host, port, database, username, password

charts/posthog/templates/_postgresql.tpl
macobo: `charts/posthog/templates/_snippet-redis-env.tpl` does exactly the opposite - it defines the password via secrets

charts/posthog/values.yaml
macobo: There's a trade-off here - do we want to keep the keys consistent with other services or consistent with `postgresql.`? 

I'd argue for `posgresql.` but ymmv.

charts/posthog/templates/_postgresql.tpl
macobo: From our call: The comment was about using a different secrets file, rather than main secrets.yaml

charts/posthog/values.yaml
hazzadous: I think I'd lean on the side of mirroring `postgresql.`, but I can see it might be surprising to see a difference to the other configs.

charts/posthog/templates/_postgresql.tpl
```diff
@@ -99,29 +97,39 @@ Set postgres port
...
+{{- if .Values.postgresql.enabled -}}
+{{- .Values.postgresql.postgresqlDatabase | quote -}}
+{{- else -}}
+{{- .Values.externalPostgresql.postgresqlDatabase | quote -}}
+{{- end -}}
 {{- end -}}
 
 {{/*
-Set postgres URL
+Set if postgres secret should be created
```
guidoiaquinti, line 129:
```suggestion
Return true if a secret object for PostgreSQL should be created
```

guidoiaquinti (approved):
Very nice cleanup

guidoiaquinti:
> @guidoiaquinti given I'm out will leave merging this to you - all good on my end :)

Yep! My plan was to do it tomorrow after sprint planning. Enjoy your vacation! No more GH and Slack! 😝

PR #286, opened 2022-02-02
fix(email): ensure email.existingSecretKey is used

hazzadous:
## Description

Previously we would ignore the values `email.existingSecretKey` and default to "smtp-password" dispite it being an option in the chart values.yaml. Here I've added some template unittesting and factored out these env settings into a snippet. See individual commits for separate tasks.

## Type of change

## How has this been tested?

I'm relying on the unittests here, and the existing k8s integration tests.

## Checklist


guidoiaquinti (approved):
LGTM

PR #288, opened 2022-02-03
feat(events): use readyz/livez for healthcheck probes

hazzadous:
## Description

This adds healthchecks specifically for the events service, with the
intention to ensure that:

 1. on initial start of event pods, they will only be marked as ready if
        all dependencies are reachable
 2. we do not go down if postgres is not reachable

Open question: how to handle docker image version compat.?

Open question: how to handle future changes on the docker image side? Ideally I'd love for the interface to just be `/_livez` and `/_readyz` across the board, and for each service to know what it should care about. Perhaps this is a failure of the `exclude` pattern ? 

## Type of change

## How has this been tested?

I need to add a test to validate that events doesn't come up if kafka is down.

## Checklist

### DIFF
--- a/charts/posthog/templates/events-deployment.yaml
+++ b/charts/posthog/templates/events-deployment.yaml
@@ -112,29 +112,28 @@ spec:
 {{ toYaml .Values.web.env | indent 8 }}
 {{- end }}
         livenessProbe:
-          failureThreshold: {{ .Values.web.livenessProbe.failureThreshold }}
           httpGet:
             path: /_livez/
             port: {{ .Values.service.internalPort }}
             scheme: HTTP
+          failureThreshold: {{ .Values.web.livenessProbe.failureThreshold }}
           initialDelaySeconds: {{ .Values.web.livenessProbe.initialDelaySeconds }}
           periodSeconds: {{ .Values.web.livenessProbe.periodSeconds }}
           successThreshold: {{ .Values.web.livenessProbe.successThreshold }}
           timeoutSeconds: {{ .Values.web.livenessProbe.timeoutSeconds }}
         readinessProbe:
-          failureThreshold: {{ .Values.web.readinessProbe.failureThreshold }}
           httpGet:
             # For readiness, we want to use the checks specific to the events
             # role, which may be a subset of all the apps dependencies
             path: /_readyz/?role=events
             port: {{ .Values.service.internalPort }}
             scheme: HTTP
+          failureThreshold: {{ .Values.web.readinessProbe.failureThreshold }}
           initialDelaySeconds: {{ .Values.web.readinessProbe.initialDelaySeconds }}
           periodSeconds: {{ .Values.web.readinessProbe.periodSeconds }}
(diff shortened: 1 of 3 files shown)


charts/posthog/templates/events-deployment.yaml
```diff
@@ -118,7 +118,7 @@ spec:
         livenessProbe:
           failureThreshold: {{ .Values.web.livenessProbe.failureThreshold }}
           httpGet:
-            path: /_health/
+            path: /_livez/
```
hazzadous, line 121: @guidoiaquinti would you mind helping me with the release process here?

This will actually not fail on old posthog images, as it redirects to login page (rather than returning a 404), but `/_health/` was testing more.

Not so relevant for liveness as I don't think it should be testing much more than returning a response, but for readiness is will be an issue.

How should this be managed?

charts/posthog/templates/events-deployment.yaml
```diff
@@ -128,13 +128,24 @@ spec:
         readinessProbe:
           failureThreshold: {{ .Values.web.readinessProbe.failureThreshold }}
           httpGet:
-            path: /_health/
+            # For readiness, we don't want to bring down a pod if it can't query
+            # postgres. I this case we want to remain up, and still push to kafka
+            path: /_readyz/?exclude=postgres&exclude=postgres_migrations_uptodate
```
hazzadous, line 133: @guidoiaquinti I thought `exclude` was an ok solution here, but I don't think it is now. Having the chart need to know the internals seems brittle.

I was thinking instead just have the chart know which service/role we are, i.e. pass in `role=events` then have the app decide what is important in this context.

charts/posthog/templates/events-deployment.yaml
guidoiaquinti: Good question. I don't think there's a clear winner here, I trust your judgment. Personally I think we should probably cut a new PH release, reference it here as default and then merge this PR.

hazzadous:
closing to clean up the PRs list, will revisit when we're testing this in battle

hazzadous:
Reopening now that we are looking at moving the events pod to EKS.

One thing that seems relevant is that the current _health endpoint appears to cause the events ECS tasks to go down when PostgreSQL isn't available. I'd like to try and make sure nothing except kafka can take down a pod in EKS.

guidoiaquinti (approved):

charts/posthog/values.yaml
```diff
@@ -121,34 +118,46 @@ web:
 
   livenessProbe:
     # -- The liveness probe failure threshold
-    failureThreshold: 5
+    failureThreshold: 3
     # -- The liveness probe initial delay seconds
-    initialDelaySeconds: 50
+    initialDelaySeconds: 0
```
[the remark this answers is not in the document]
hazzadous: So from my understanding the liveness/readiness probes will run after the startupProbe has succeeded.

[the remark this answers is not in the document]
guidoiaquinti: True, you've added the startupProbe in this PR 🤦

PR #289, opened 2022-02-04
Monitoring: cleanup statsd service & config

guidoiaquinti:
## Description
I've started this PR to simply cleanup and re-organize the current `statsd` service configuration. Unfortunately it ended up being much more as we've found out the `metrics.enabled:true` install a semi-copy-paste-version of the stack we could simply get from using the `statsd` upstream chart. Not sure why this was added in the first place but it looks like a useless duplication of resources and code to me. 
 

1. reorganised chart dependencies in `Chart.yaml` (and ran `helm dependency update` to regenerate the lock file)
1. removed 2 useless `namespace` entry in `Chart.yaml`
1. moved the `prometheus-statsd-exporter` chart from using the alias `statsd` to the default (we will add more exporter in the future and this will avoid confusion).
1. centralised `statsd` env variables + added unit test for it 
1. added the possibility to use an external `statsd` service by using the `externalStatsd` variable
1. removed all the resources installed via `monitor.enabled`

Related release notes:  https://github.com/PostHog/posthog.com/pull/2915

## Type of change

## How has this been tested?
CI is ✅ 

## Checklist


charts/posthog/tests/_statsd.yaml
```diff
@@ -0,0 +1,82 @@
+# Tests charts/posthog/templates/_statsd.tpl
+suite: _statsd.tpl template
+templates:
+  - templates/test-templates.yaml
```
hazzadous, line 4: What are these test templates? Can we not just check that env vars are in the env array?

hazzadous: I e.g. would this work, and be more concrete:

```
    asserts:
      - contains:
          path: spec.template.spec.containers[0].env
          content:
            name: STATSD_HOST
            value: RELEASE-NAME-posthog-prometheus-statsd-exporter
```

etc.

hazzadous: (we can then also verify that the envs are added to the appropriate deployments)

hazzadous (changes requested):
just the tests comment, I feel like we should be able to test that the envs are set in the right places

charts/posthog/tests/_statsd.yaml
guidoiaquinti: Sure, I can add that

charts/posthog/templates/test-templates.yaml
```diff
@@ -6,5 +6,7 @@ data:
 {{- include "snippet.clickhouse-env" . | nindent 2 }}
 {{- else if .Values.testTemplates.postgresqlEnvTest -}}
 {{- include "snippet.postgresql-migrate-env" . | nindent 2 }}
+{{- else if .Values.testTemplates.prometheusStatsdExporterEnvTest -}}
```
hazzadous, line 9: I'm not quite sure what the point of this template is, do we need it? It seems like something we're adding to test snippets implementation detail rather than the chart itself.

charts/posthog/tests/events-deployment.yaml
```diff
@@ -32,3 +32,63 @@ tests:
           count: 1
       - isKind:
           of: Deployment
+
+  - it: spec.template.spec.containers[0].env should have the default statsd env variables when `prometheus-statsd-exporter.enabled` is `true`
```
hazzadous, line 36: If the tests for the deployments are the same, would it be better to just define them once and test across all deployments? Or is this too rigid?

charts/posthog/templates/test-templates.yaml
guidoiaquinti: This was added by Karl to test variable templates that might not end being rendered. In this case it is probably overkill.

charts/posthog/tests/events-deployment.yaml
guidoiaquinti: Ok, let me change it

hazzadous (approved):

PR #291, opened 2022-02-07
Monitoring - add `prometheus-kafka-exporter` & upgrade Prometheus chart

guidoiaquinti:
## Description

* add `prometheus-kafka-exporter` as dependency (optional)
* upgrade the prometheus upstream chart to the latest version (optional) 

## Type of change

## How has this been tested?
CI is ✅ 
Tested locally

## Checklist


charts/posthog/values.yaml
```diff
@@ -745,6 +746,33 @@ externalStatsd:
...
+###
+###
+### ---- prometheus-kafka-exporter ----
+###
+###
+prometheus-kafka-exporter:
+  # -- Whether to install the `prometheus-kafka-exporter` or not.
+  enabled: false
+
+  # -- We want to pin to image tag `v1.4.2` as it is currently the only
```
hazzadous, line 758: Is this holding us back a version or pinning past the detail?

charts/posthog/values.yaml
```diff
@@ -745,6 +746,33 @@ externalStatsd:
...
+  # TODO: remove the override once `prometheus-kafka-exporter` will default to this version.
+  #
+  image:
+    tag: v1.4.2
+
+  # -- Map of annotations to add to the pods.
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/path: /metrics
+    prometheus.io/port: "9308"
```
hazzadous, line 770: Where does this port number come from. I can see it is used as the default service port, but is it guaranteed to be the pod target port also?

hazzadous (approved):
Looks good

Just some questions around whether CI would pick up issues, non-blocking though

charts/posthog/values.yaml
guidoiaquinti: Nope, I've simply set this to the latest release in order to be able to test this locally on my mac

charts/posthog/values.yaml
```diff
@@ -745,6 +746,33 @@ externalStatsd:
...
+
+  # -- Specify the target Kafka brokers to monitor.
+  kafkaServer:
+    - posthog-posthog-kafka:9092
```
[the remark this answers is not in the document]
guidoiaquinti: > I guess we're relying on this matching up with how we configure internal kafka

Yes

> is there anything in CI that will tell us if this is wrong?

No but I'm also not sure we should (? I can add a test if you want)

charts/posthog/values.yaml
guidoiaquinti: The port number is coming from the exporter binary itself

charts/posthog/values.yaml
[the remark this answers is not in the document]
hazzadous: We should validate the correctness to some extent, if that's just that the kubetest deployment wouldn't end up in a ready state I think that would be enough.

I wouldn't want the only time that this is run to be 1. local dev 2. running in anger by someone downstream.

hazzadous (approved):

PR #292, opened 2022-02-07
Monitoring - add prometheus-redis-exporter

guidoiaquinti:
## Description

## Type of change

## How has this been tested?

## Checklist


charts/posthog/values.yaml
```diff
@@ -773,6 +773,52 @@ prometheus-kafka-exporter:
...
+    prometheus.io/scrape: "true"
+    prometheus.io/path: /metrics
+    prometheus.io/port: "9187"
+
+  # -- Configuration options.
+  config:
+    datasource:
+      host: posthog-posthog-postgresql
+      user: postgres
+      passwordSecret: {
```
hazzadous, line 797: non-blocking 

What's with the `{}` btw?

charts/posthog/values.yaml
```diff
@@ -773,6 +773,52 @@ prometheus-kafka-exporter:
...
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/path: /metrics
+    prometheus.io/port: "9187"
+
+  # -- Configuration options.
+  config:
+    datasource:
+      host: posthog-posthog-postgresql
+      user: postgres
```
hazzadous, line 796: Do we offer setting this in the helm chart to something that isn't `postgres`?

hazzadous (approved):
Assuming CI passes it looks good to go.

Just someunimportant comments about yaml syntax

charts/posthog/values.yaml
guidoiaquinti: It is valid YAML but I can remove it if you want.

charts/posthog/values.yaml
guidoiaquinti: I'm not sure but I can remove it to simply default to https://github.com/helm/charts/blob/390ee6614b4bb0da96b397f9b382e3ddca9f59a4/stable/prometheus-postgres-exporter/values.yaml#L66

guidoiaquinti: I was keeping it to make clear here's a user/password block

PR #298, opened 2022-02-14
CI - change k8s version for DO

guidoiaquinti:
## Description
See title. We need to change it as the previous one is not available anymore:

```
doctl k8s clusters create \
    helm-test-e2e-do-install-5c418b0 \
    --version 1.21.5-do.0 \
    --tag="provisioned_by:github_action" \
    --size s-2vcpu-4gb \
    --count 2 \
    --wait
  shell: /usr/bin/bash -e {0}
Error: POST https://api.digitalocean.com/v2/kubernetes/clusters: 422 (request "9a[3](https://github.com/PostHog/charts-clickhouse/runs/5182026377?check_suite_focus=true#step:6:3)d9592-ae15-[4](https://github.com/PostHog/charts-clickhouse/runs/5182026377?check_suite_focus=true#step:6:4)bc2-8fff-[6](https://github.com/PostHog/charts-clickhouse/runs/5182026377?check_suite_focus=true#step:6:6)cad931[7](https://github.com/PostHog/charts-clickhouse/runs/5182026377?check_suite_focus=true#step:6:7)f72b") validation error: invalid version slug
```

DO changelog is available [here](https://docs.digitalocean.com/products/kubernetes/changelog/).

## Type of change

## How has this been tested?
Tested locally.

## Checklist

### DIFF
--- a/.github/workflows/test-digitalocean-install.yaml
+++ b/.github/workflows/test-digitalocean-install.yaml
@@ -40,7 +40,7 @@ jobs:
       run: |
         doctl k8s clusters create \
           ${{ steps.vars.outputs.k8s_cluster_name }} \
-          --version 1.21.5-do.0 \
+          --version 1.21.9-do.0 \
           --tag="provisioned_by:github_action" \
           --size s-2vcpu-4gb \
           --count 2 \


tiina303 (approved):
the benefit of hardcoding the version is  that things don't break underneath us unexpectedly? 
Or why can't we just drop this from here & look into it if the test starts breaking?

guidoiaquinti:
> the benefit of hardcoding the version is that things don't break underneath us unexpectedly?
> Or why can't we just drop this from here & look into it if the test starts breaking?

Yes, we would like to test a specific version and we should come up with a better way of picking it.

PR #313, opened 2022-02-25
Kafka - fix for `externalKafka.brokers`

guidoiaquinti:
## Description
Make `externalKafka.brokers` a list and build `KAFKA_HOSTS` and `KAFKA_URL` out of the items in there. Unfortunately we don't use the same name as well as format for those two environment variables:

- `KAFKA_HOSTS` is used by the `plugin-server` and expects a string of comma-separated items in the form "host1:port1"
- `KAFKA_URL` is used by PostHog/web and expects a string of comma separated items in the form "kafka://host1:port1"

As unifying the names and format in `posthog/posthog` will require to wait for a release, let's abstract this by handing the differences directly in the Helm chart. 

## Type of change
Note: we are adding bump major to this PR so we will release a new major version of the chart including this change and #271.

## How has this been tested?
CI should be passing ✅ 

## Checklist

### DIFF
--- a/charts/posthog/templates/_kafka.tpl
+++ b/charts/posthog/templates/_kafka.tpl
@@ -31,14 +31,14 @@
 {{- end }}
 - name: KAFKA_ENABLED
   value: "true"
-# Used by PostHog/plugin-server. Expected format: comma-separated list of "host:port"
+# Used by PostHog/plugin-server. There is no specific reason for the difference. Expected format: comma-separated list of "host:port"
 - name: KAFKA_HOSTS
 {{- if .Values.kafka.enabled }}
   value: {{ ( include "posthog.kafka.brokers" . ) }}
 {{ else }}
   value: {{ join "," .Values.externalKafka.brokers | quote }}
 {{- end }}
-# Used by PostHog/web. Expected format: comma-separated list of "kafka://host:port"
+# Used by PostHog/web. There is no specific reason for the difference. Expected format: comma-separated list of "kafka://host:port"
 - name: KAFKA_URL
 {{- if .Values.kafka.enabled }}
   value: {{ printf "kafka://%s" ( include "posthog.kafka.brokers" . ) }}
--- a/charts/posthog/templates/_kafka.tpl
+++ b/charts/posthog/templates/_kafka.tpl
@@ -1,31 +1,48 @@
 {{/* Common Kafka ENV variables and helpers used by PostHog */}}
 
 {{/* Return the Kafka fullname */}}
-{{- define "posthog.kafka.fullname" -}}
-{{- if .Values.kafka.fullnameOverride -}}
-{{- .Values.kafka.fullnameOverride | trunc 63 | trimSuffix "-" -}}
-{{- else if .Values.kafka.nameOverride -}}
-{{- printf "%s-%s" .Release.Name .Values.kafka.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- define "posthog.kafka.fullname" }}
+{{- if .Values.kafka.fullnameOverride }}
+{{- .Values.kafka.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else if .Values.kafka.nameOverride }}
+{{- printf "%s-%s" .Release.Name .Values.kafka.nameOverride | trunc 63 | trimSuffix "-" }}
 {{- else -}}
-{{- printf "%s-%s" (include "posthog.fullname" .) "kafka" | trunc 63 | trimSuffix "-" -}}
-{{- end -}}
-{{- end -}}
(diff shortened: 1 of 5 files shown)


charts/posthog/values.yaml
```diff
@@ -485,8 +485,8 @@ kafka:
       - posthog-posthog-zookeeper:2181
 
 externalKafka:
-  # - External Kafka brokers. Ignored if `kafka.enabled` is set to `true`. Multiple brokers can be provided in a comma separated list, e.g. `host1:port1,host2:port2`.
-  brokers: ""
+  # - External Kafka brokers. Ignored if `kafka.enabled` is set to `true`. Multiple brokers can be provided as array/list.
```
guidoiaquinti, line 488: This is not a breaking change as we didn't cut a release (see https://github.com/PostHog/charts-clickhouse/pull/271)

charts/posthog/templates/test-templates.yaml
```diff
@@ -6,6 +6,8 @@ data:
 {{- include "snippet.clickhouse-env" . | nindent 2 }}
 {{- else if .Values.testTemplates.postgresqlEnvTest -}}
 {{- include "snippet.postgresql-migrate-env" . | nindent 2 }}
+{{- else if .Values.testTemplates.kafkaEnvTest -}}
```
hazzadous, line 9: Do we need this? Unless it’s used in process of development I’d propose removing

charts/posthog/templates/_kafka.tpl
```diff
@@ -1,31 +1,48 @@
...
-  value: "kafka://{{ include "posthog.kafka.brokers" . }}"
+
+{{- $hostsWithPrefix := list }}
+{{- range $host := .Values.externalKafka.brokers }}
+{{- $hostWithPrefix := (printf "kafka://%s" $host) }}
+{{- $hostsWithPrefix = append $hostsWithPrefix $hostWithPrefix }}
+{{- end }}
 - name: KAFKA_ENABLED
   value: "true"
+# Used by PostHog/plugin-server. Expected format: comma-separated list of "host:port"
```
hazzadous, line 34: Can we explicitly say there is no specific reason for the difference? Otherwise this appears at first site as “this is the way it is”, and you need to dig through pr/commit messages to understand

hazzadous (approved):
Looks good, a couple of comments but no blocking

charts/posthog/templates/_kafka.tpl
guidoiaquinti: You mean remove the comments entirely?

hazzadous: No actually state that there is no specific reason for the difference

hazzadous: But is very minor, just adds context for future people who aren’t us. But honestly no need

charts/posthog/templates/_kafka.tpl
```diff
@@ -1,31 +1,48 @@
 {{/* Common Kafka ENV variables and helpers used by PostHog */}}
 
 {{/* Return the Kafka fullname */}}
-{{- define "posthog.kafka.fullname" -}}
-{{- if .Values.kafka.fullnameOverride -}}
-{{- .Values.kafka.fullnameOverride | trunc 63 | trimSuffix "-" -}}
-{{- else if .Values.kafka.nameOverride -}}
-{{- printf "%s-%s" .Release.Name .Values.kafka.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- define "posthog.kafka.fullname" }}
```
macobo, line 4: Q: Why not remove the whitespace here?

This is a thing I've been thinking about the more I've worked with helm - we should start standardizing the whitespace approaches. There's too many to count.

charts/posthog/templates/_kafka.tpl
```diff
@@ -1,31 +1,48 @@
...
+{{- end }}
 
 {{/* ENV used by PostHog deployments for connecting to Kafka */}}
+
 {{- define "snippet.kafka-env" }}
-- name: KAFKA_HOSTS
-  value: "{{ include "posthog.kafka.brokers" . }}"
-- name: KAFKA_URL
-  value: "kafka://{{ include "posthog.kafka.brokers" . }}"
+
```
macobo, line 26: Q: Why the blank line here? If it's to avoid issues with whitepsace, `nindent` seems like a better fit.

charts/posthog/templates/_kafka.tpl
guidoiaquinti: Good question! Do you have a proposal about which standard should we use?

macobo: I think I do (always strip whitespace + use nindent). I'll create a proposal PR soonish cleaning up a few files! :)

PR #319, opened 2022-03-04
Raise errors if both externalService and service are set

### LINKED ISSUE #318: Raise an error if externalService is used without service.enabled being false
### Proposed change

If `service.enabled` is true, `externalService` values are ignored. This can cause errors due to typos.

We should raise errors for example `externalClickhouse.host` is set but `clickhouse.host` is not.

### Alternative options

Let users debug

## Additional context
Ran into this doing https://github.com/PostHog/charts-clickhouse/issues/279 - typos are inevitable!

We can use a similar system to our current backwards incompatibility errors.

a3n7:
## Description
Fixes #318 

## Type of change

## How has this been tested?
```shell
kind create cluster --name posthog
kubectx kind-posthog

# fails as expected
helm upgrade --install posthog-test \
	~/git/charts-clickhouse/charts/posthog \
	--set cloud=private \
	--set postgresql.enabled=true \
	--set externalPostgresql.postgresqlHost=test123.local

# works
helm upgrade --install posthog-test \
	~/git/charts-clickhouse/charts/posthog \
	--set cloud=private \
	--set postgresql.enabled=true \
	--set externalPostgresql.postgresqlHost=""
```

## Checklist


charts/posthog/templates/error-on-both-externalService-and-service-enabled.yaml
```diff
@@ -0,0 +1,5 @@
+{{- if false -}}
+# :TRICKY: This template raises an error if any externalService is configured without service.enabled being false in values.yaml.
+#   It doesn't actually define any real kubernetes resources
+{{- end -}}
+{{- include "snippet.error-on-both-externalService-and-service-enabled" . -}}
```
macobo, line 5: Minor request: please merge this file with `charts/posthog/templates/error-on-removed-values.yaml` to `charts/posthog/templates/error-on-invalid-values.yaml`, merge the tpl file with the other tpl file.

After that this looks great to me. It would be awesome to add some helm-unittest coverage here, but not required immediately.

a3n7: Done. I couldn't add an error test case because of [this bug](https://github.com/quintush/helm-unittest/issues/144), but it should be possible to do so once the bug fix is released.

macobo:
This is great! Thank you for the contribution :bow:

PR #321, opened 2022-03-08
feat(async-migrations): Check async migrations state

tiina303:
## Description

For https://github.com/PostHog/posthog/issues/8691

The plan is to make migrate job fail and all pods that run code to have async migration check too. Will also be making 1.34 a major upgrade and mention that async migrations need to be done beforehand in the [upgrade notes](https://posthog.com/docs/self-host/deploy/upgrade-notes)

`/bin/migrate-check` is available on 1.33 already https://github.com/PostHog/posthog/blob/1.33.0/bin/migrate-check
however `bin/migrate` needs to be changed & released as we don't want to be running the async migrations (https://github.com/PostHog/posthog/blob/1.33.0/bin/migrate). The fix is in https://github.com/PostHog/posthog/pull/8872 so after that we can start using `/bin/migrate` in migrate job.

I'm not sure how this was expected to work before, but adding `set -e` was one of the options to make things work as we need - i.e. fail out when a line failed. I updated 2 locations, we use bin/sh more and I'm not sure anything needs to change for web/other pods where it's just a single cmd.
With code like this 
```
        command:
          - /bin/sh
          - -c
          - |
            echo "failing the next line"
            fail
            echo "shouldn't reach here"
```
the logs in migrate job were & migrate job completed successfully
```
failing the next line                                                                                                               
/bin/sh: fail: not found                                                                                                            
shouldn't reach here
Stream closed EOF for posthog/posthog-migrate-2022-03-08-20-17-15-ztfjj (migrate-job)
```

With code like this
```
        command:
          - /bin/sh
          - -c
          - |
            set -e
            echo "failing the next line"
            fail
            echo "shouldn't reach here"

```
Logs were & migrate job errored 
```
failing the next line                                                                                                               
/bin/sh: fail: not found                                                                                                            
Stream closed EOF for posthog/posthog-migrate-2022-03-08-20-17-15-8czzx (migrate-job)
```

## Type of change

## How has this been tested?

Ran `helm upgrade --reuse-values posthog ~/Posthog/charts-clickhouse/charts/posthog` which worked fine. Checked migrate job and events precheck output to confirm we checked async migrations.

Upgraded the migrate job & containers init to have `fail` as a line mimicing async migrations check failing & ran
```
helm upgrade --reuse-values posthog ~/Posthog/charts-clickhouse/charts/posthog  --atomic --wait --wait-for-jobs --debug --timeout=2m
```
Got the following output, which unfortunately isn't all that great, but I haven't been able to figure out a better solution
```
wait.go:48: [debug] beginning wait for 53 resources with timeout of 2m0s
ready.go:277: [debug] Deployment is not ready: posthog/posthog-events. 0 out of 1 expected pods are ready
ready.go:277: [debug] Deployment is not ready: posthog/posthog-events. 0 out of 1 expected pods are ready
...
ready.go:277: [debug] Deployment is not ready: posthog/posthog-events. 0 out of 1 expected pods are ready
upgrade.go:433: [debug] warning: Upgrade "posthog" failed: rate: Wait(n=1) would exceed context deadline
upgrade.go:451: [debug] Upgrade failed and atomic is set, rolling back to last successful release
history.go:56: [debug] getting history for release posthog
rollback.go:65: [debug] preparing rollback of posthog
rollback.go:113: [debug] rolling back posthog (current: v29, target: v28)
rollback.go:72: [debug] creating rolled back release for posthog
rollback.go:78: [debug] performing rollback of posthog
...
Error: UPGRADE FAILED: release posthog failed, and has been rolled back due to atomic being set: rate: Wait(n=1) would exceed context deadline
helm.go:84: [debug] rate: Wait(n=1) would exceed context deadline
release posthog failed, and has been rolled back due to atomic being set
```
viewing things in k9s confirms that the old pods for events etc stayed up and the update was rolled back, now down-time.
Screen Shot 2022-03-08 at 20 36 21

Unfortunately the error messages in the console aren't very helpful and since we roll back the migrate jobs get deleted too and we just see a completed one (that was spun up for the rollback).

I couldn't figure out any way to access the migrate job logs later either, we can do 
```
kubectl get events -n posthog -o wide --sort-by=.metadata.creationTimestamp | grep migrate
```
and see that there were various migrate pods, but can't access their logs. Nor see much useful info in the output from above.

## Checklist

### DIFF
--- a/charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
+++ b/charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
@@ -5,8 +5,9 @@
   command:
     - /bin/sh
     - -c
-    - >
-        until (python manage.py migrate --check);
+    - |
+        set -e
+        until (bin/migrate-check);
         do
             echo "Waiting for database migrations to be completed"; sleep 1;
         done
--- a/charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
+++ b/charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
@@ -5,9 +5,8 @@
   command:
     - /bin/sh
     - -c
-    - |
-        set -e
-        until (bin/migrate-check);
+    - >
+        until (python manage.py migrate --check);
         do
             echo "Waiting for database migrations to be completed"; sleep 1;
         done
--- a/charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
+++ b/charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
@@ -5,8 +5,7 @@
   command:
     - /bin/sh
     - -c
-    - |
-        set -e
+    - >
         {{ if .Values.clickhouse.enabled }}
         until (nc -vz "{{ include "posthog.clickhouse.fullname" . }}.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local" 8123);
         do
--- a/charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
+++ b/charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
@@ -5,7 +5,8 @@
   command:
     - /bin/sh
     - -c
-    - >
+    - |
+        set -e
         {{ if .Values.clickhouse.enabled }}
         until (nc -vz "{{ include "posthog.clickhouse.fullname" . }}.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local" 8123);
         do
--- a/charts/posthog/templates/migrate.job.yaml
+++ b/charts/posthog/templates/migrate.job.yaml
@@ -33,12 +33,16 @@ spec:
       containers:
       - name: migrate-job
         image: {{ template "posthog.image.fullPath" . }}
+        # switch over to using /bin/migrate once 1.34 is released - https://github.com/PostHog/posthog/pull/8872
         command:
           - /bin/sh
           - -c
           - |
+            set -e
             python manage.py notify_helm_install || true
-            python manage.py migrate && python manage.py migrate_clickhouse
+            python manage.py migrate
+            python manage.py migrate_clickhouse
+            python manage.py run_async_migrations --check
         env:
         # Kafka env variables
         {{- include "snippet.kafka-env" . | indent 8 }}


charts/posthog/templates/migrate.job.yaml
```diff
@@ -33,12 +33,16 @@ spec:
...
         command:
           - /bin/sh
           - -c
           - |
+            set -e
             python manage.py notify_helm_install || true
-            python manage.py migrate && python manage.py migrate_clickhouse
+            python manage.py migrate
+            python manage.py migrate_clickhouse
+            python manage.py run_async_migrations --check
```
hazzadous, line 45: What's the deal with only having `--check` here where as the others don't? Worth adding a comment with the reasoning as at first glance if looks like a mistake.

hazzadous: Is there another place that the migrations are actually run? If not then I think we'll be in a deadlock if we have migrations to apply, i.e. init container will never complete successfully

hazzadous (commented):
Not quite sure how the migrations are started, do we need to start them via the UI?

charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
```diff
@@ -5,8 +5,9 @@
   command:
     - /bin/sh
     - -c
-    - >
-        until (python manage.py migrate --check);
+    - |
+        set -e
```
guidoiaquinti, line 9: If we `set -e` then the until become useless

charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
```diff
@@ -5,7 +5,8 @@
   command:
     - /bin/sh
     - -c
-    - >
+    - |
+        set -e
```
guidoiaquinti, line 9: Ditto

charts/posthog/templates/migrate.job.yaml
tiina303: Async migrations can take a long time (e.g. 5h) which is why we don't want to run them during the helm upgrade. They need to be ran beforehand. If they haven't been completed we'd see an error (it will also be a major upgrade note if they need to be completed). We do need to run the django & CH migrations before to (1) be able to run the check (2) some of the async migrations would be auto-completed (note we do auto-complete them if we can within `--check`). If you have better ideas how to do this I'm all ears.

tiina303: Yes, the task would wait forever & end up doing a rollback if `--atomic` was used (see description)

charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
tiina303: not sure what's the best way, but if there's any other problems here they will just be bypassed (see description)

guidoiaquinti: This is an init container to verify service dependency (e.g. is ClickHouse up and running?)

I don't get why we are adding a `set -e` here but maybe I've misunderstood something

tiina303: If you look at the description "I'm not sure how this was expected to work before, but adding set -e ..." until "Type of change" does that explain it? I can remove it from here, but if we make a typo in some line it would just be ignored and look as all is well.

charts/posthog/templates/migrate.job.yaml
```diff
@@ -33,12 +33,16 @@ spec:
       containers:
       - name: migrate-job
         image: {{ template "posthog.image.fullPath" . }}
+        # switch over to using /bin/migrate once 1.34 is released - https://github.com/PostHog/posthog/pull/8872
         command:
           - /bin/sh
           - -c
           - |
+            set -e
```
tiina303, line 41: we do definitely need it here based on description (or something equivalent)

charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
guidoiaquinti: If we `set -e` and ClickHouse is not ready the script instead of looping will `exit 1`

tiina303: why would it be `exit 1` in that case? It's the same outcome if I have 
```
set -e
until (false);
do 
  echo "with set -e"
done
```

as 
```
until (false);
do 
  echo "without set -e"
done
```
As in I saw the echo output from the init container

guidoiaquinti: My bad, I thought it was propagated to the subshell

tiina303: I'll let you decide if you want to have it anywhere else, shouldn't be part of this PR anyway

tiina303:
Is this good to ship or are there any more open questions?

tiina303:
CI failure seems to be https://github.com/PostHog/charts-clickhouse/pull/327

PR #341, opened 2022-03-30
chore(MinIO): add (optional) object storage service

guidoiaquinti:
## Description
This PR adds opt-in object storage service capabilities to the Helm chart.

**Why?**
We are working to add some new features to PostHog that will rely on an object storage service. Without it those features will not be enabled 

Documentation change is available here https://github.com/PostHog/posthog.com/pull/3263

## Type of change

## How has this been tested?
1️⃣ Manual testing:
* the service is provisioned and it is healthy
* PostHog pods gets the expected environment variables

2️⃣ CI is ✅ 

## Checklist


charts/posthog/tests/object-storage-settings.yaml
```diff
@@ -0,0 +1,168 @@
+suite: Object storage settings propagated to pods
+templates:
+  # This list may be too broad or too narrow, I've just added to all deployments
+  # that use the django backend.
+  - templates/web-deployment.yaml
```
hazzadous, line 5: missing plugins?

charts/posthog/values.yaml
```diff
@@ -736,6 +736,47 @@ serviceAccount:
...
+    rootUser: root-user
+    # -- MinIO root password
+    rootPassword: root-password-change-me-please
+    # -- Use existing secret for credentials details (`auth.rootUser` and `auth.rootPassword` will be ignored and picked up from this secret). The secret has to contain the keys `root-user` and `root-password`)
+    existingSecret:
+  persistence:
+    # -- Enable MinIO data persistence using PVC.
+    enabled: true
+  # -- Comma, semi-colon or space separated list of buckets to create at initialization (only in standalone mode)
+  defaultBuckets: "posthog"
```
hazzadous, line 758: I don't see anything in the test for this

hazzadous (approved):
Looks good

A kubetest might not be amiss also. I guess we are waiting on landing the code in `posthog` repo first before getting this in?

guidoiaquinti:
> A kubetest might not be amiss also. I guess we are waiting on landing the code in posthog repo first before getting this in?

If we want something e2e yes. Otherwise I could simply test if the API endpoint is there and credential works. I think we could extend the current PostHog django & plugin-server healthcheck to check and then rely here on the HTTP code

charts/posthog/values.yaml
```diff
@@ -736,6 +736,56 @@ serviceAccount:
...
+      api: "19000"
+      # -- MinIO Console service port
+      console: "19001"
+
+  ## -- MinIO pod(s) annotation.
+  podAnnotations:
+    # Uncomment those lines if you want Prometheus server to scrape MinIO pods metrics.
+    # prometheus.io/scrape: "true"
+    # prometheus.io/path: "/minio/v2/metrics/cluster"
+    # prometheus.io/port: "9000"
```
hazzadous, line 775: Should this be 19000 as well?

guidoiaquinti: No it is the POD port so 9000 works

PR #346, opened 2022-03-31
chore(ci): explicitly set reclaim policy for volumes

guidoiaquinti:
## Description
Explicitly set the reclaim policy of the default `StorageClass` to `Delete` to make sure we don't keep disk volumes around after cluster deletion.

## Type of change

## How has this been tested?
CI is ✅ 

## Checklist

### DIFF
--- a/.github/workflows/test-amazon-web-services-install.yaml
+++ b/.github/workflows/test-amazon-web-services-install.yaml
@@ -76,7 +76,8 @@ jobs:
 
         # Note: we explicitly set the reclaim policy of the default StorageClass
         # to 'Delete' to make sure we don't keep disk volumes around after cluster deletion
-        kubectl patch storageclass -p '{"reclaimPolicy": "Delete"}'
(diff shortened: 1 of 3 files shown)


hazzadous (approved):
Looks good

Have kicked off the workflows 🤞

guidoiaquinti:
I forgot we run those on `main` only. Looks like they are failing 👀

guidoiaquinti:
Mmmm I think this PR is useless:

- `storageclass.storage.k8s.io/gp2 patched (no change)`
- `storageclass.storage.k8s.io/do-block-storage patched (no change)`
- `storageclass.storage.k8s.io/standard patched (no change)`

Looking at the documentation:

> GKE might not delete the following resources:
> 
> - [External load balancers](https://cloud.google.com/compute/docs/load-balancing) created by the cluster
> - [Internal load balancers](https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing) created by the cluster
> - [Persistent disk](https://cloud.google.com/compute/docs/disks#pdspecs) volumes
> 
> Ref: https://cloud.google.com/kubernetes-engine/docs/how-to/deleting-a-cluster

and similarly on AWS.

I think the best approach here would be running a Helm uninstall to remove the resources and then deleting a cluster. I'm closing this without merging. Sorry for the spam.

PR #349, opened 2022-04-05
fix(ingress-nginx): forward X-Forwarded-* headers to upstreams

hazzadous:
## Description

Within the posthog app we use the X-Forwarded-* headers to, for
instance, determine the scheme to use when building absolute urls.
Previously we were not setting X-Forwarded-* headers hence we ended up
producing http scheme urls, which causes issue, among other things, with
cross domain requests.

Here we set on the nginx ingress controller that we should always
forward these headers. We could also have added these as an annotation
to the ingress resources as a snippet, but I've just gone for applying
this across the board.

This was as a result of debugging
https://posthog.slack.com/archives/C02G72Y6E9H/p1646072061024549

## Type of change

## How has this been tested?

tl;dr I haven' added any automated tests here to avoid making them dependent on implementation details, checking for absolute urls in specific api endpoints doesn't fill me with joy.

I've triggered builds for [AWS](https://github.com/PostHog/charts-clickhouse/actions/runs/2097585084), [DO](https://github.com/PostHog/charts-clickhouse/actions/runs/2102936237), [GCP](https://github.com/PostHog/charts-clickhouse/actions/runs/2102939703). Waiting for them to complete.

AWS completed successfully which I'm inclined to say that means use-forwarded-headers doesn't interfere with use-proxy-protocol that is specified in the e2e AWS values file.

I was thinking there must be something around secure cookies/csrf tokens that I'd be able to add as a test. The cookies at least are all about browser enforcement so that doesn't work. There are mentions of csrf token depending on is_secure.

There are some bits of the app that use Django's `request.build_absolute_url` which is actually the issue that initiated this change, but I don't want to assume that much detail.

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -325,6 +325,15 @@ ingress:
   # -- TLS secret to be used by the ingress.
   secretName:
 
+ingress-nginx:
+  controller:
+    config:
+      # -- Whether to forward "X-Forwarded-*" headers to upstream services.
+      # -- This is needed to ensure the PostHog application knows e.g. if the
+      # -- downstream proxy is using a secure connection. See the official
+      # -- [ingress-nginx documentation](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#use-forwarded-headers)
+      use-forwarded-headers: "true"
+
 ###
 ###
 ### ---- POSTGRESQL ----


charts/posthog/values.yaml
```diff
@@ -325,6 +325,12 @@ ingress:
   # -- TLS secret to be used by the ingress.
   secretName:
 
+ingress-nginx:
```
guidoiaquinti, line 328: Nit: I think we've been using 2 spaces for indentation

guidoiaquinti:
LGTM but we should probably:

- make sure they play well with [those other annotations](https://posthog.com/docs/self-host/deploy/aws) (same for DO,...)
- make sure we update the docs ☝️

guidoiaquinti:
I think you might need to add it as well to our `values.yaml` files used in CI ((example](https://github.com/PostHog/charts-clickhouse/blob/main/ci/values/amazon_web_services.yaml#L27-L42))

PR #358, opened 2022-04-26
Make email for ACME optional in values

fuziontech:
## Description
Make using `james.g@posthog.com` optional as the ACME email address associated with cert manager

## Type of change

## How has this been tested?
Locally on home cluster box

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -274,9 +274,9 @@ cert-manager:
   # -- Whether to install `cert-manager` CRDs.
   installCRDs: true
   # -- Who to email if the certificate is about to expire
-  # -- Defaults to James.g@posthog.com but feel free (please)
+  # -- Defaults to `Null` but feel free (please)
   # -- to replace this with whoever is managing the stack
-  email: james.g@posthog.com 
+  email: None 
 
   #
   # [Workaround] - do not use the local DNS for the 'cert-manager' pods since it would return local IPs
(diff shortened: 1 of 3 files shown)


charts/posthog/values.yaml
```diff
@@ -273,6 +273,10 @@ cert-manager:
   enabled: false
   # -- Whether to install `cert-manager` CRDs.
   installCRDs: true
+  # -- Who to email if the certificate is about to expire
+  # -- Defaults to James.g@posthog.com but feel free (please)
+  # -- to replace this with whoever is managing the stack
+  email: james.g@posthog.com
```
hazzadous, line 279: It feels messy for a posthog uses instance to end up triggering an email going to your work email address, this feels like a potential attack vector (not that you are malicious yourself of course! But we shouldn't have access to someone else's communications. I'm sure people will be not setting this.)

Instead can we make this default to empty? What's the behaviour then?

tiina303 (approved):
It's an improvement though, so fine with 🚢  and iterate

guidoiaquinti:
> It's an improvement though, so fine with 🚢 and iterate

True but we are changing a public API here, iterating on it is not very easy and it's a disruptive change for our end users.

charts/posthog/values.yaml
fuziontech: 100% Updating with something we might be able to use in the future.

PR #363, opened 2022-05-03
feat: Add redis memory policy

tiina303:
## Description

Similar to https://github.com/PostHog/posthog/pull/9573 
Except on the chart we want to set it to close to the default disk size, but with buffer because ([here](https://redis.io/docs/manual/eviction/#how-the-eviction-process-works))
> we continuously cross the boundaries of the memory limit, by going over it, and then by evicting keys to return back under the limits.

## Type of change

## How has this been tested?

Installed locally with 1-click default values file & local charts repo, jumped into the redis pod to check mem config "(note that before it was 0 and noevict)
```
$ redis-cli
> info memory
...
maxmemory:4718592000
maxmemory_human:4.39G
maxmemory_policy:allkeys-lru
...
```
It's possible to override in values (updated having set extraFlags to have 200mb) & verified that I saw the corresponding maxmemory value.

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -449,7 +449,7 @@ redis:
       ## The maxmemory configuration directive is used in order to configure Redis to use a specified
       ## amount of memory for the data set. Setting maxmemory to zero results into no memory limits
       ## see https://redis.io/topics/lru-cache for more details
-      - "--maxmemory 4500mb"
+      - "--maxmemory 200mb"
       ## The exact behavior Redis follows when the maxmemory limit is reached is configured using the
       ## maxmemory-policy configuration directive
       ## allkeys-lru: evict keys by trying to remove the less recently used (LRU) keys first, in order
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -445,6 +445,7 @@ redis:
 
       # -- Persistent Volume size.
       size: 5Gi
+    # -- Array with additional command line flags for Redis master.
     extraFlags:
       ## The maxmemory configuration directive is used in order to configure Redis to use a specified
       ## amount of memory for the data set. Setting maxmemory to zero results into no memory limits
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -450,7 +450,7 @@ redis:
       ## The maxmemory configuration directive is used in order to configure Redis to use a specified
       ## amount of memory for the data set. Setting maxmemory to zero results into no memory limits
       ## see https://redis.io/topics/lru-cache for more details
-      - "--maxmemory 200mb"
+      - "--maxmemory 400mb"
       ## The exact behavior Redis follows when the maxmemory limit is reached is configured using the
       ## maxmemory-policy configuration directive
       ## allkeys-lru: evict keys by trying to remove the less recently used (LRU) keys first, in order
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -439,6 +439,10 @@ redis:
     existingSecretPasswordKey: ""
   
   commonConfiguration: |-
+      # Enable AOF https://redis.io/topics/persistence#append-only-file
+      appendonly yes
+      # Disable RDB persistence, AOF persistence already enabled.
+      save ""
       ## The maxmemory configuration directive is used in order to configure Redis to use a specified
       ## amount of memory for the data set. Setting maxmemory to zero results into no memory limits
       ## see https://redis.io/topics/lru-cache for more details
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -445,6 +445,16 @@ redis:
 
       # -- Persistent Volume size.
       size: 5Gi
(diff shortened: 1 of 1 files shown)


charts/posthog/values.yaml
```diff
@@ -445,6 +445,16 @@ redis:
 
       # -- Persistent Volume size.
       size: 5Gi
+    extraFlags:
```
guidoiaquinti, line 448:
```suggestion
    # -- Array with additional command line flags for Redis master
    extraFlags:
```

guidoiaquinti:
Should we add this as `redis.commonConfiguration` instead of `redis.master.extraFlags`? Replica nodes will not pick this information otherwise.

guidoiaquinti:
LGTM @tiina303 can you please only double check if the new config gets merged with [this](https://github.com/bitnami/charts/blob/ad6a41df9de2204da957f54b9309a0a4f8e04fd5/bitnami/redis/values.yaml#L132-L139) or not? I would like to don't lose those settings

tiina303:
Good call, it didn't specifically saw `aof_enabled:0` in the info output after this change vs before it was 1 & now that I copied those values it's 1 again. Is this good like this or do we want to use the flags approach I had before because if bitnami were to change the default we wouldn't get that change now.

tiina303:
> Also, do you know what's the impact on the app when Redis is not available? Technically ingestion should not be impacted but the UI might be. Can you please verify it? Thank you! 🙇

The plugin-server healthcheck uses redis, though Yakko is changing that in https://github.com/PostHog/posthog/pull/9581 so plugin-server would show up not healthy.

guidoiaquinti:
>Is this good like this or do we want to use the flags approach I had before because if bitnami were to change the default we wouldn't get that change now.

I wasn't aware of this limitation either. Due to this, I think your initial approach was good, apologies for having asked you to change it 🙈

charts/posthog/values.yaml
```diff
@@ -445,6 +445,16 @@ redis:
...
+      ## The maxmemory configuration directive is used in order to configure Redis to use a specified
+      ## amount of memory for the data set. Setting maxmemory to zero results into no memory limits
+      ## see https://redis.io/topics/lru-cache for more details
+      - "--maxmemory 4500mb"
```
guidoiaquinti, line 452: Didn't we use 200MB earlier? Otherwise it will be a quite costly service to run. Do we really need that much memory? 🤔

tiina303: We used 200MB for hobby + docker-compose. Here I set it to 4500MB based on the default for Redis disk we ask is 5GB

tiina303: yeah exactly I just realized that too 🤦‍♀️  why do we need 5GB of disk for Redis, what is it used for?
Will set it to 200MB like everywhere else unless we think it should be something else.

tiina303, line 452:
```suggestion
      - "--maxmemory 200mb"
```

yakkomajuri: I do think 200mb is too low, however. This will probably blow up a lot of instances like N****ta and V*****ta that use the default Redis.

Worth asking them how much they use? One of them we have k8s access anyway. 

Also we should not merge this before the Snowflake stuff and whatever else needs changing.

yakkomajuri: I can tell you V. has 3gb available on Redis 360mb

yakkomajuri: Instance status shows these

tiina303: that's cool, added the maxmemory info there too as that's relevant too now https://github.com/PostHog/posthog/pull/9660

tiina303: Do we want to set it to be relatively small like everything else (the current chart is rather small default based & I'd like to keep it that way so small enterprises can get off the ground more easily) and make it major upgrade asking bigger users like V. to pick their limit?

yakkomajuri: Well my bigger concern is breaking existing users

yakkomajuri: If we can minimize the number of users we break on self-hosted it would be great.

Let's maybe do 400mb?

tiina303: sounds good to me

charts/posthog/values.yaml
```diff
@@ -445,6 +445,17 @@ redis:
 
       # -- Persistent Volume size.
       size: 5Gi
+    # -- Array with additional command line flags for Redis master.
+    extraFlags:
+      ## The maxmemory configuration directive is used in order to configure Redis to use a specified
+      ## amount of memory for the data set. Setting maxmemory to zero results into no memory limits
+      ## see https://redis.io/topics/lru-cache for more details
+      - "--maxmemory 200mb"
```
tiina303, line 453:
```suggestion
      - "--maxmemory 400mb"
```

fuziontech (approved):
`400mb` and `allkeys-lru` seems reasonable to me.

I know this is for a dependency chart, but would be nice to see if we could add tests here as well.

PR #366, opened 2022-05-04
Remove `/bin/sh -c` from deployment commands

joetrollo:
## Description
Many deployments in the chart started containers using `/bin/sh -c` despite only running a single command and not a multi-line script. This PR switches those deployments to run the commands directly, which is both simpler and eliminates the issue of propagating SIGTERM signals to the applications (as discovered in https://github.com/PostHog/posthog/issues/9440).

## Type of change

## How has this been tested?
I deployed PostHog to my cluster with these changes and verified that all the deployments started normally and successfully.

## Checklist


guidoiaquinti (approved):
LGTM. Thank you for your contribution!

PR #371, opened 2022-05-10
fix(clickhouse-operator): pin docker image version

hazzadous:
Previously we were using the latest available image. Here we pin to the
latest image.

It looks like they version the chart and the operator image the same. It
may be worth pinning them to the same version. However, this will cause
the operator to downgrade in existing installs, so here I'm opting to
try to keep the image version the same as current installs.

## Type of change

## Checklist


guidoiaquinti:
I think you need to update the test snapshots but overall LGTM. Thank you!

guidoiaquinti (approved):

PR #373, opened 2022-05-10
test(clickhouse): wait on desired state, not intermediary

hazzadous:
Previously we were using kubernetes details to decide if ClickHouse had
scaled up correctly. This appeared to fail quite regularly though, so
instead I am using one ClickHouse properties to decide if we converge on
the desired state.

## Type of change

## Checklist

### DIFF
--- a/ci/kubetest/test_clickhouse_sharding.py
+++ b/ci/kubetest/test_clickhouse_sharding.py
@@ -1,4 +1,5 @@
 import pytest
+import time
 import yaml
 
 from helpers.clickhouse import get_clickhouse_table_counts_on_all_nodes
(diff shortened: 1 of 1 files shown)


ci/kubetest/test_clickhouse_sharding.py
```diff
@@ -48,17 +50,15 @@ def test_upgrading_to_more_shards(kube):
...
-            "clickhouse.altinity.com/chi": "posthog",
-        },
-        expected_count=6,
-    )
-
-    # Validate the ClickHouse node setup
-    number_of_hosts, table_counts = get_clickhouse_table_counts_on_all_nodes(kube)
+
+    # Wait for new replicas to come up and for tables to be created
+    for attempt in range(24):
```
guidoiaquinti, line 55: Nit: I usually prefer to set a deadline so that the overall timeframe is fixed

```
start = time.time()
while time.time() - start < something: # e.g. 10 minutes
```

guidoiaquinti (approved):

PR #376, opened 2022-05-10
clickhouse-operator: upgrade from 0.16.1 to 0.18.4

guidoiaquinti:
## Description
Upgrade the `clickhouse-operator` from version `0.16.1` to `0.18.4`. This brings some reliability and security fixes from upstream (full changelog is available [here](https://github.com/Altinity/clickhouse-operator/releases)).

As we can see from the diff, there are few CRD changes that we need to considered as breaking changes on our end. Release notes are available [here](https://github.com/PostHog/posthog.com/pull/3421). 

## Type of change

## How has this been tested?
1️⃣  CI is ✅ 

2️⃣  manually tested the upgrade path locally:

* new installation: ✅ 
* old installation (from `main`) -> upgrade (from this branch): CH pod doesn't get restarted, only the operator pod (as expected) ✅  
* old installation (from `main`) -> upgrade (from this branch) without following the release notes: CH pod doesn't get restarted, only the operator pod (as expected) ✅  

## Checklist

### DIFF
--- a/scripts/clickhouse_operator_sync.sh
+++ b/scripts/clickhouse_operator_sync.sh
@@ -18,7 +18,7 @@ TMP_FOLDER="$(mktemp -d)"
 trap 'rm -rf -- "$TMP_FOLDER"' EXIT
 
 OPERATOR_NAMESPACE="posthog"
-CLICKHOUSE_OPERATOR_TAG="0.16.1"
+CLICKHOUSE_OPERATOR_TAG="0.18.4"
 URL="https://raw.githubusercontent.com/Altinity/clickhouse-operator/${CLICKHOUSE_OPERATOR_TAG}/deploy/operator/clickhouse-operator-install-template.yaml"
 
 #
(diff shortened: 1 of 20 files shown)


scripts/clickhouse_operator_sync.sh
```diff
@@ -62,16 +62,16 @@ kubectl-slice -f "$TMP_FOLDER/clickhouse-operator.yaml" -o "${CHART_PATH}/templa
 FILES="${CHART_PATH}/templates/clickhouse-operator/*"
 for f in $FILES
 do
-    sed -i'' '1i\
```
guidoiaquinti, line 65: This is a fix for a regression that was recently introduced via #371 (the regression is fixed on macOS) cc @hazzadous

hazzadous: This breaks gnu sed support I believe

guidoiaquinti: TIL https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux

PR #406, opened 2022-06-01
add support for kubernetes 1.24

jredl-va:
## Description
We can no longer template the posthog helm chart as our kubernetes version passed that of the support release. Bumping kubernetes support to 1.24


guidoiaquinti:
> If its only CI thats blocking this, can we add the CI support back in when 1.24 is available for your tool chain? 

Yes we can.

> Nothing about the latest kubernetes release should affect posthog and posthog dependencies?

That's a question we usually use CI to validate :)

guidoiaquinti:
k8s is now supported by the chart via #421. Closing this

PR #431, opened 2022-06-13
feat: add ability to specify existing secret for Django SECRET_KEY

hazzadous:
Previously the behaviour was to look for an existing value for the
secret and, if one doesn't exist, to generate one automatically.

There wasn't a way to explicitly specify the key, which means that if
you want to point a new deployment at existing database resources,
you'll end up invalidating e.g. Django issued cookies.

## Type of change

## Checklist


guidoiaquinti (approved):
Overall LGTM but what about removing the reference to Django? With that we could have a more generic `posthog-env`

hazzadous:
> Overall LGTM but what about removing the reference to Django? With that we could have a more generic `posthog-env`

Yup wasn't sure about that either, will update.

charts/posthog/templates/migrate.job.yaml
```diff
@@ -66,11 +66,8 @@ spec:
           value: {{ template "posthog.site.url" . }}
         - name: DEPLOYMENT
           value: {{ template "posthog.deploymentEnv" . }}
-        - name: SECRET_KEY
-          valueFrom:
-            secretKeyRef:
-              name: {{ template "posthog.fullname" . }}
-              key: posthog-secret
+        # Django env variables
```
guidoiaquinti, line 69:
```suggestion
  # PostHog app settings
```

charts/posthog/templates/secrets.yaml
```diff
@@ -9,6 +9,7 @@ type: Opaque
 data:
   {{- if $previous }}
   posthog-secret: {{ index $previous.data "posthog-secret" }}
+  {{- else if .Values.posthogSecretKey.existingSecret }}
```
guidoiaquinti, line 12: is this empty?

charts/posthog/templates/web-deployment.yaml
```diff
@@ -90,11 +90,8 @@ spec:
         - name: SECURE_COOKIES
           value: '0'
       {{- end }}
-        - name: SECRET_KEY
-          valueFrom:
-            secretKeyRef:
-              name: {{ template "posthog.fullname" . }}
-              key: posthog-secret
+        # Django env variables
```
guidoiaquinti, line 93:
```suggestion
  # PostHog app settings
```

PR #434, opened 2022-06-13
chore: remove unique-name secret

hazzadous:
It doesn't appear to be being used.

As a result of adding the ability to specify your own `SECRET_KEY` via
`posthogSecretKey.existingKey` it is now the case that if you also set
`email.existingKey` your'll end up with `data == null` within the
`posthog` secret. This causes issues with the evaluation of `index
$previous.data 'unique-name'`

Rather than resolving this, I've opted for removal due to it not being
used.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/secrets.yaml
+++ b/charts/posthog/templates/secrets.yaml
@@ -17,16 +17,3 @@ data:
   {{- if not .Values.email.existingSecret }}
   smtp-password: {{ default "" .Values.email.password | b64enc | quote }}
   {{- end }}
-
-  {{- if $previous }}
-  # :TRICKY: Raise an error if trying to apply the incorrect values file for clients we manage under client_values/
-  {{- $prevUniqueName := b64dec (default "" (index $previous.data "unique-name")) }}
-  {{- $uniqueName := default "" .Values.uniqueName }}
-  {{- if and (ne $uniqueName $prevUniqueName) (ne $prevUniqueName "") }}
-  {{ fail "uniqueName does not match expected! Are you using the wrong values.yaml?" }}
-  {{- end }}
-  {{- end }}
-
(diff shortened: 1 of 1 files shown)


guidoiaquinti (approved):
Nice 🧹

charts/posthog/templates/secrets.yaml
```diff
@@ -17,16 +17,3 @@ data:
   {{- if not .Values.email.existingSecret }}
   smtp-password: {{ default "" .Values.email.password | b64enc | quote }}
   {{- end }}
-
```
guidoiaquinti, line 20: Do we need the `$previous` above? Can't we make it idempotent without the lookup?

hazzadous: It's still used to carry across the posthog-secret between runs

hazzadous: Is there another way to handle this?

guidoiaquinti: Maybe there are some scenarios where removing it can trigger a breaking change but if our users either provide credentials as k8s secrets or as a plaintext value in `values.yaml`, this lookup should not be necessary.

hazzadous: Hmm yes, the update is the tricky bit! I'm going to leave as is for now if that's ok.

guidoiaquinti: Sure 👍

hazzadous:
@tiina303 do you happen to know if this is still needed? I ask only because github recommended you as a reviewer

hazzadous:
Going to yolo this in, given the unlikely failures in the pipeline being related 🙄 and handle them separately.

PR #438, opened 2022-06-14
Plugin server split

macobo:
## Description

This PR introduces `pluginsAsync` - a separate plugin server process that allows scaling non-ingestion pieces of plugin-server separately from everything else. 

Note that this isn't turned on by default as it's:
- Experimental - the exact service boundaries are bound to change over time (as the number of these split processes)
- We'll handle running only the main process gracefully in the chart and generally

Note I also added an E2E test for measuring `onEvent` plugin calls - in a split plugin server, these are handled separately.

## Type of change

## How has this been tested?

- Test deploy
- Unit tests
- E2E ingestion test.

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -180,7 +180,7 @@ worker:
 
 plugins:
   # -- Whether to install the PostHog plugin-server stack or not.
-  # This service handles data ingestion into clickhouse, running apps and async jobs.
+  # This service handles data ingestion into ClickHouse, running apps and async jobs.
   # See `pluginsAsync` to scale this separately.
   enabled: true
 
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -179,10 +179,12 @@ worker:
     enabled: false
 
 plugins:
-   # -- Whether to install the PostHog plugin-server stack or not.
+  # -- Whether to install the PostHog plugin-server stack or not.
+  # This service handles data ingestion into clickhouse, running apps and async jobs.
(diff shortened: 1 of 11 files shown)


charts/posthog/values.yaml
```diff
@@ -179,10 +179,12 @@ worker:
     enabled: false
 
 plugins:
-   # -- Whether to install the PostHog plugin-server stack or not.
+  # -- Whether to install the PostHog plugin-server stack or not.
+  # This service handles data ingestion into clickhouse, running apps and async jobs.
```
guidoiaquinti, line 183:
```suggestion
  # This service handles data ingestion into ClickHouse, running apps and async jobs.
```

ci/values/k3s-plugin-server-split.yaml
```diff
@@ -0,0 +1,38 @@
+cloud: k3s
+
+image:
+  # :TODO: Remove this once 1.37.0 is out
+  tag: latest
```
guidoiaquinti, line 5: Can we use a sha here to reduce any CI flakiness might be created by PostHog images?

ci/values/k3s.yaml
```diff
@@ -9,6 +9,9 @@ ingress:
 web:
   secureCookies: false
 
+internalMetrics:
```
guidoiaquinti, line 12: wrong indentation? It should be `web.internalMetrics` right now it seems to be at the top level

macobo: Thanks! This caused me trouble without knowing about this :sweat_smile:

PR #439, opened 2022-06-15
ClickHouse - switch default service to ClusterIP

guidoiaquinti:
## Description
This PR switch the default ClickHouse service type from `NodePort` to `ClusterIP`. This is to remove the possibility of exposing the service in environments where k8s nodes are not deployed in private subnets (or in public subnets but without network restriction in place).

## Type of change

## How has this been tested?
CI should be ✅ 

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -676,8 +676,8 @@ clickhouse:
     runAsGroup: 101
     fsGroup: 101
 
-  # -- Service Type: LoadBalancer (allows external access) or NodePort (more secure, no extra cost)
-  serviceType: NodePort
+  # -- Kubernetes Service type.
+  serviceType: ClusterIP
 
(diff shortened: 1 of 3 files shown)


charts/posthog/values.yaml
```diff
@@ -676,8 +676,15 @@ clickhouse:
     runAsGroup: 101
     fsGroup: 101
 
-  # -- Service Type: LoadBalancer (allows external access) or NodePort (more secure, no extra cost)
-  serviceType: NodePort
+  # -- Kubernetes Service type.
+  serviceType: ClusterIP
+
+  # -- A whitelist of IP addresses or network masks the ClickHouse user is
```
guidoiaquinti, line 682:
```suggestion
  # -- An allowlist of IP addresses or network masks the ClickHouse user is
```

PR #444, opened 2022-06-23
feat(hpa): add non-abstracted hpa config values

hazzadous:
Previously we had a thin abstraction over a subset of the hpa config.
Rather than maintain this abstraction I have simply exposed the
underlying spec for hpa manifests. This was we can use the full range of
hpa without needing to make chart interface changes.

I've added this as the existing configuration can exhibit a lot of
flapping, where we might go from 10 pods, down to 5, only to then
request another 5 pods directly after. I would like to set the
`behavior.scaleDown.stabilizationWindowSeconds` setting to a higher
value than the default.

Note that this does not include the pluginsAsync values addition at the
moment. This is because we do not have an HPA manifest for pluginsAsync,
I'll handle this separately.

The reason for not reusing e.g. `web.hpa` is to avoid any unexpected
behaviour with existing values people may have. I chose `autoscaling`
bsaed on a similarly structured value in the `ingress-nginx` helm chart.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/events-hpa.yaml
+++ b/charts/posthog/templates/events-hpa.yaml
@@ -9,7 +9,11 @@ spec:
     kind: Deployment
     apiVersion: apps/v1
     name: {{ template "posthog.fullname" . }}-events
+  {{ if .Values.events.autoscaling }}
+  {{ .Values.events.autoscaling | nindent 2 }}
+  {{ else }}
   minReplicas: {{ .Values.events.hpa.minpods }}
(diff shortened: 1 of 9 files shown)


guidoiaquinti (approved):

hazzadous:
It looks like the test failure is related to disk pressure and unrelated to this change. @guidoiaquinti I'm going to merge this and have a look at reducing disk usage separately.

PR #446, opened 2022-06-23
fix(hpa): specify behavior as annotation

hazzadous:
The `behavior` is not part of the `autoscaling/v1` spec, but rather part
of `autoscaling/v2beta2` and `autoscaling/v2`. Rather than change to
`autoscaling/v2beta2` for which I need to check the update path, I'm
proposing to add as an annotation.

@guidoiaquinti I'm not entirely sure what "The new fields introduced in autoscaling/v2beta2 are preserved as annotations when working with autoscaling/v1" means in https://v1-21.docs.kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/ , perhaps you or @benjackwhite have a better view here.

I'm also happy to update this PR to use `autoscaling/v2beta2` if you think that's a better approach.

For context, [this](https://github.com/PostHog/posthog-cloud-infra/pull/302) is the change I want to be able to make.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/events-hpa.yaml
+++ b/charts/posthog/templates/events-hpa.yaml
@@ -4,6 +4,8 @@ kind: HorizontalPodAutoscaler
 metadata:
   name: {{ template "posthog.fullname" . }}-events
   labels: {{- include "_snippet-metadata-labels-common" . | nindent 4 }}
+  annotations:
+    autoscaling.alpha.kubernetes.io/behavior: {{ toJson .Values.events.hpa.behavior | quote }}
 spec:
   scaleTargetRef:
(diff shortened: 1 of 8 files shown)


benjackwhite (commented):
If you have tested this and it works then two thumbs up. To me though the description reads more like "If you set call the Kube API requesting v1, we will take the behaviour attributes and append them as annotations" rather than "you can use annotations for this feature if you aren't using v2".

Not super deep into it though... As I say, if it works then awesome, otherwise might be worth doing a check for version support (akin to [something like this](https://renehernandez.io/snippets/customize-api-version-helm-templates/) )

PR #450, opened 2022-06-27
feat(events): allow event pod resources to be set independently

hazzadous:
Previously they just came from the `web.resources` setting. Now you can:

 1. set `events.resources` for use by events pods
 2. update helm chart version, only changing the behaviour if you
    explicitly set the `events.resources` value. Otherwise, you should
    get the existing behaviour of using the `web.resources` value.

## Type of change

## Checklist


hazzadous:
> CI is failing otherwise LGTM

Will need to debug the plugin server split one further, I think it fails even with the multistage image build, perhaps it's spewing some logs out that are causing the disk space issues. But I think unrelated to this PR

PR #454, opened 2022-06-28
chore: ensure we can deploy multiple releases in separate namespaces

hazzadous:
This is partially to ensure we can run tests in clean envs, but is also
relevant to uses e.g. wanting to run prod and staging in the same
cluster.

Previously ClusterRoles and ClusterRoleBindings for the
clickhouse-operator were not unique, this uses the namespace to correct
that.

## Type of change

## Checklist


guidoiaquinti:
This PR ended up being more than a couple of changes. Here's a list of what has been done:

* refactor the current `pytest` parallelism as each test is now independent from others
* made `clusterrole` and `clusterrolebinding` coming from the `clickhouse-operator` unique across the namespace
* added in `conftest` logic to automatically wipe both k8s and helm before the start of each test
* minor fixes to black/isort/flake8 config
* disabled few flaky/broken tests (like clickhouse sharding, clickhouse backup, and #462). I will fix those in a follow up PR but there's no point to block the pipeline for those

PR #457, opened 2022-06-29
fix: wait for clickhouse migrations before starting new pods

hazzadous:
Previously we were just waiting on PostgreSQL migrations. As a result,
pods would end up getting into a crash loop.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
+++ b/charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
@@ -8,7 +8,12 @@
     - >
         until (python manage.py migrate --check);
         do
-            echo "Waiting for database migrations to be completed"; sleep 1;
+            echo "Waiting for PostgreSQL database migrations to be completed"; sleep 1;
+        done
+
+        until (python manage.py migrate_clickhouse --check);
+        do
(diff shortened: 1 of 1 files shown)


charts/posthog/templates/_snippet-initContainers-wait-for-migrations.tpl
```diff
@@ -8,7 +8,12 @@
     - >
         until (python manage.py migrate --check);
         do
-            echo "Waiting for database migrations to be completed"; sleep 1;
+            echo "Waiting for PostgreSQL database migrations to be completed"; sleep 1;
+        done
+
+        until (python manage.py migrate_clickhouse --check);
+        do
+            echo "Waiting for ClickHouse database migrations to be completed"; sleep 1;
```
guidoiaquinti, line 16: LGTM, should we add as well a loop for `python manage.py run_async_migrations --check` and `python manage.py sync_replicated_schema`?

hazzadous: Maybe, but I'd rather get those in separately if that's ok? I don't feel super comfortable on what the functionality should be for those ones.

guidoiaquinti (approved):

hazzadous:
I'll roll the restart pr into one

PR #458, opened 2022-06-29
ci: detect restarted pods in kubetest

hazzadous:
Before https://github.com/PostHog/charts-clickhouse/pull/457/checks we
were having pods getting into crash loops. To ensure we catch these
cases in the future, and to enable some faster failing tests we abort as
soon as we see something has restarted.

## Type of change

## Checklist


guidoiaquinti (approved):
LGTM. Let's see if this gets ✅ because I think the current condig is not good enough to avoid restarts.

PR #468, opened 2022-07-01
deps(loki): use the simple scalable Helm Chart instead

hazzadous:
This chart splits read and write into separate pods. I'm doing this to
give better isolation between read and write workloads and the handle
larger scale of logs.

The values for this are not plug and play with the plain loki chart, and require migrating existing config under `structuredConfig`

## Type of change

## Checklist


guidoiaquinti (approved):
Are the `values.yaml` values plug-n-play in this version? If yes, LGTM

hazzadous:
> Are the `values.yaml` values plug-n-play in this version? If yes, LGTM

Maybe not for the config, may need to add them under structuredConfig but I just pushed up to see what CI had to say. I'll change to draft

hazzadous:
Closing to clean up the list. Can reopen when we look at e.g. query performance increases.

PR #471, opened 2022-07-04
ci(kubetest): remove unnecessary calls to the k8s API

hazzadous:
We are getting calls failing when calling the `status()` method of the
kubetest API, due to either intermittent failures or pods disappearing.
We don't need the restart anyhow as we have all the info we need from
`get_pods`.

I also:

 1. run a job per test file to remove some layers of indirection
 2. add some retries for fetching some ClickHouse resources
 3. splits up some test files to ensure they run in parallel

## Type of change

## Checklist


hazzadous:
Assuming the tests pass with the latest commit @guidoiaquinti I'll merge this in. It looks like the remaining failures are timeouts waiting for the migrations to run. Fingers crossed squashing the migrations is a simple task.

PR #473, opened 2022-07-05
PgBouncer - upgrade to 1.17.0

guidoiaquinti:
## Description
- upgrade PgBouncer from 1.12.0 to 1.17.0 (see changelog [here](https://www.pgbouncer.org/changelog.html#pgbouncer-112x)). We are doing that by upgrading the default image to an official bitnami one. From the changelog I don't see any major breaking change.   
- add more fine grain readiness/liveness checks + lifecycle config
- various other fixes to align this deployment and service more close to the others we already have 

Release notes are available at: https://github.com/PostHog/posthog.com/pull/3717

## Type of change
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## How has this been tested?
CI ✅ 

## Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #477, opened 2022-07-06
PgBouncer - use pkill instead of killall

guidoiaquinti:
## Description
Small but big detail (introduced via #473), the `bitnami/pgbouncer` container image doesn't contain the `killall` binary, let's use `pkill` instead:

```
I have no name!@posthog-pgbouncer-556c59c447-vxxpk:/$ killall -INT pgbouncer
bash: killall: command not found
```

```
I have no name!@posthog-pgbouncer-556c59c447-vxxpk:/$ /usr/bin/pkill --version
pkill from procps-ng 3.3.17
```

## Type of change

## How has this been tested?
CI is ✅ + manual tests of the command

## Checklist

### DIFF
--- a/charts/posthog/templates/pgbouncer-deployment.yaml
+++ b/charts/posthog/templates/pgbouncer-deployment.yaml
@@ -108,7 +108,7 @@ spec:
               #
               # See https://www.pgbouncer.org/usage.html
               #
-              command: ["/bin/sh", "-c", "killall -INT pgbouncer && sleep 60"]
+              command: ["/bin/sh", "-c", "pkill -INT pgbouncer && sleep 60"]
 
       {{- if .Values.pgbouncer.extraVolumeMounts }}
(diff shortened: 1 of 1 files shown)


hazzadous (approved):

charts/posthog/templates/pgbouncer-deployment.yaml
```diff
@@ -108,7 +108,7 @@ spec:
               #
               # See https://www.pgbouncer.org/usage.html
               #
-              command: ["/bin/sh", "-c", "killall -INT pgbouncer && sleep 60"]
+              command: ["/bin/sh", "-c", "pkill -INT pgbouncer && sleep 60"]
```
hazzadous, line 111: @guidoiaquinti I wonder if the sleep is unnecessary? Presumably pgbouncer will exit at some point and the pod will be removed at that point? If it doesn't exit within the grace period it will be hard killed

PR #479, opened 2022-07-07
feat(pgbouncer): add Prometheus exporter as sidecar

hazzadous:
If enabled we deploy along side each PGBouncer instance an exporter. See
https://github.com/prometheus-community/pgbouncer_exporter for details.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/pgbouncer-deployment.yaml
+++ b/charts/posthog/templates/pgbouncer-deployment.yaml
@@ -81,11 +81,6 @@ spec:
         - name: PGBOUNCER_DATABASE
           value: {{ include "posthog.postgresql.database" . }}
 
-        {{ if .Values.pgbouncer.exporter.enabled }}
-        - name: PGBOUNCER_STATS_USERS
-          value: exporter
-        {{ end }}
-
 {{- if .Values.pgbouncer.env }}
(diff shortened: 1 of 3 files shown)


charts/posthog/templates/pgbouncer-deployment.yaml
```diff
@@ -21,6 +21,10 @@ spec:
         {{- if .Values.pgbouncer.podAnnotations }}
 {{ toYaml .Values.pgbouncer.podAnnotations | indent 8 }}
         {{- end }}
+
+        prometheus.io/scrape: "true"
```
guidoiaquinti, line 25: I think this should go as container annotation, not pod annotation?

charts/posthog/values.yaml
```diff
@@ -581,7 +597,9 @@ pgbouncer:
     - name: PGBOUNCER_MAX_CLIENT_CONN
       value: "1000"
     - name: PGBOUNCER_POOL_MODE
-      value: "transaction"
+      value: transaction
+    - name: PGBOUNCER_IGNORE_STARTUP_PARAMETERS
+      value: extra_float_digits
```
guidoiaquinti, line 602: I've added this as the exporter requires it. Looking at our codebase I don't think we rely on this setting anywhere.

PR #497, opened 2022-07-26
Make sentryDSN configurable for plugin server

macobo:
## Description

After this change, sentryDSN is changable for plugin server.  This matches how we want to do tracking on cloud.

Opted for this route (over env vars) so new deployments always receive _a_ SentryDSN.

## Type of change

## How has this been tested?

See tests.

## Checklist

### DIFF
--- a/charts/posthog/templates/_posthog.tpl
+++ b/charts/posthog/templates/_posthog.tpl
@@ -16,12 +16,15 @@
       key: {{ .Values.posthogSecretKey.existingSecretKey }}
 - name: SITE_URL
   value: {{ template "posthog.site.url" . }}
-- name: SENTRY_DSN
-  value: {{ .Values.sentryDSN | quote }}
 - name: DEPLOYMENT
   value: {{ template "posthog.deploymentEnv" . }}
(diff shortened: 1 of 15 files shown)


hazzadous (approved):

guidoiaquinti (approved):
Personally I don't like having implementation specific references in something that should be generic. I think we should leverage default + deployment specific ENV config but I also understand your concern.

macobo:
> I think we should leverage default + deployment specific ENV config but I also understand your concern.

I would love to do that - but how? The problem is that helm blows up with a very obscure error if there's two env vars with the same name and there's no way to avoid that AFAIK.

guidoiaquinti:
> > I think we should leverage default + deployment specific ENV config but I also understand your concern.
> 
> I would love to do that - but how? The problem is that helm blows up with a very obscure error if there's two env vars with the same name and there's no way to avoid that AFAIK.

True, that's not a great UX but we could probably wrap those around some Helm merge function (`merge, `mustMerge`, `mergeOverwrite` ...) anyway a bigger task for the context of this PR

PR #499, opened 2022-07-27
chore(metrics): hide /_metrics endpoint

hazzadous:
To ensure we do not leak data on private paths, we direct these paths to
a non-existent backend. There is almost certainly a better way to do
this, but this is my straw man.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -53,6 +53,18 @@ spec:
         paths:
           {{- if (ne (include "ingress.type" .) "clb") }}
           - pathType: Prefix
+            path: "/_metrics"
+          {{- else }}
+          - pathType: ImplementationSpecific
+            path: "/_metrics"
+          {{- end }}
+            backend:
+              service:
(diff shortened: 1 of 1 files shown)


charts/posthog/templates/ingress.yaml
```diff
@@ -52,6 +52,18 @@ spec:
   {{- end }}
         paths:
           {{- if (ne (include "ingress.type" .) "clb") }}
+          - pathType: Prefix
```
guidoiaquinti, line 55: What about the `clb` implementation? Specifically for NGINX we could probably leverage something like the `nginx.ingress.kubernetes.io/server-snippet` annotation (see [docs](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#server-snippet)) and do something like:

```
    nginx.ingress.kubernetes.io/server-snippet: |
      location ~* "^/_metrics" {
          deny all;
          return 403;
        }
```

hazzadous:
I'm going to close this and instead put it on a different port

PR #507, opened 2022-08-06
Fix an issue that checks for connectivity in the init container even when pgbouncer is disabled

Ozarklake:
## Description
Fix an issue that checks for connectivity in the init container even when pgbouncer is disabled, Also in snippet.postgresql-env always use the address of pgbouncer.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/_postgresql.tpl
+++ b/charts/posthog/templates/_postgresql.tpl
@@ -2,10 +2,18 @@
 
 {{/* ENV used by posthog deployments for connecting to postgresql */}}
 {{- define "snippet.postgresql-env" }}
+  {{ if .Values.pgbouncer.enabled }}
 - name: POSTHOG_POSTGRES_HOST
   value: {{ template "posthog.pgbouncer.host" . }}
 - name: POSTHOG_POSTGRES_PORT
   value: {{ include "posthog.pgbouncer.port" . | quote }}
+  {{ else }}
+- name: POSTHOG_POSTGRES_HOST
+  value: {{ template "posthog.postgresql.host" . }}
+- name: POSTHOG_POSTGRES_PORT
+  value: {{ include "posthog.postgresql.port" . | quote }}
+  {{- end }}
+
 - name: POSTHOG_DB_USER
   value: {{ include "posthog.postgresql.username" . }}
 - name: POSTHOG_DB_NAME
--- a/charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
+++ b/charts/posthog/templates/_snippet-initContainers-wait-for-service-dependencies.tpl
@@ -22,10 +22,12 @@
         done
         {{ end }}
 
+        {{ if .Values.pgbouncer.enabled }}
         until (nc -vz "{{ include "posthog.pgbouncer.host" . }}.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local" {{ include "posthog.pgbouncer.port" . }});
         do
             echo "waiting for PgBouncer"; sleep 1;
         done
+        {{ end }}
 
         {{ if .Values.postgresql.enabled }}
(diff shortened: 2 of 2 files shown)


hazzadous (approved):
Looks good

Ideally we'd have a kubetest for this also.

guidoiaquinti:
> Looks good
> 
> Ideally we'd have a kubetest for this also.

Overall I agree but you could have find this regression only by clicking in the UI and run a _test_ on the datasource. Not a trivial kubetest 😅

hazzadous:
@Ozarklake to push a little more on this, do you mind explaining a little more about what your use case is?

@macobo highlights that this might open us/you up to issues that you may not have anticipated.

Ozarklake:
> 

@hazzadous It's simple, since `value.yaml` provides some way to disable pgbouncer, than we should make things work as expected

hazzadous:
> @hazzadous It's simple, since value.yaml provides some way to disable pgbouncer, than we should make things work as expected

Yes I agree that given it's possible to disable PGBouncer in the chart then it should be possible to make it a working configuration.

The issue is that, given the way we are hitting PostgreSQL, it's likely that at scale you and others that run without PGBouncer will hit difficult to debug issues, and we'd rather keep that risk at a minimum unless there is a clear use case where running without PGBouncer makes sense.

Ozarklake:
> Yes I agree that given it's possible to disable PGBouncer in the chart then it should be possible to make it a working configuration.
> 
> The issue is that, given the way we are hitting PostgreSQL, it's likely that at scale you and others that run without PGBouncer will hit difficult to debug issues, and we'd rather keep that risk at a minimum unless there is a clear use case where running without PGBouncer makes sense.

@hazzadous  I got you now.  I'm not actually an expert on PostgreSQL or Pgbouncer, I just found that Chart wasn't working as well as I expected when I tried to deploy it myself, and after checking I realized it was probably a bug

hazzadous:
> If, in your experience, using Pgbouncer is the best option in either case, then I suggest not offering the option to turn off Pgbouncer to avoid friction

💯 yes I think this is the thing to do. Luckily with this PGBouncer bug no one will be using this option successfully #feature. I think the ideal solution here would be to offer an `externalPgbouncer` option that allows specifying e.g. AWS RDS Proxy, and either pgbouncer or externalPgbouncer needs to be set. i.e. removing the possibility of not using it.

But in the meantime yes removing `pgbouncer.enabled` is probably a sensible pragmatic solution.

I'm going to close this issue without merging, are you still able to do whatever you were trying to do?

guidoiaquinti:
> > If, in your experience, using Pgbouncer is the best option in either case, then I suggest not offering the option to turn off Pgbouncer to avoid friction
> 
> 💯 yes I think this is the thing to do. Luckily with this PGBouncer bug no one will be using this option successfully #feature. I think the ideal solution here would be to offer an `externalPgbouncer` option that allows specifying e.g. AWS RDS Proxy, and either pgbouncer or externalPgbouncer needs to be set. i.e. removing the possibility of not using it.
> 
> But in the meantime yes removing `pgbouncer.enabled` is probably a sensible pragmatic solution.
> 
> I'm going to close this issue without merging, are you still able to do whatever you were trying to do?

I think we added the possibility to disable `pgbouncer` semi-recently and only for testing purposes. I'm not against merging this PR, I think we should just make clear in the `values.yaml` with something like "_Please modify this setting only if you are really sure about what you are doing_" (I think we have other examples in the file)

Ozarklake:
> > If, in your experience, using Pgbouncer is the best option in either case, then I suggest not offering the option to turn off Pgbouncer to avoid friction
> 
> 💯 yes I think this is the thing to do. Luckily with this PGBouncer bug no one will be using this option successfully #feature. I think the ideal solution here would be to offer an `externalPgbouncer` option that allows specifying e.g. AWS RDS Proxy, and either pgbouncer or externalPgbouncer needs to be set. i.e. removing the possibility of not using it.
> 
> But in the meantime yes removing `pgbouncer.enabled` is probably a sensible pragmatic solution.
> 
> I'm going to close this issue without merging, are you still able to do whatever you were trying to do?

Sure, It doesn't affect me that

PR #530, opened 2022-08-17
Update kafka exporter chart

ellie:
I was looking into a warning I was getting while trying to deploy to a
local cluster:

```
coalesce.go:220: warning: cannot overwrite table with non table for prometheus-kafka-exporter.kafka.sidecars (map[])
```

Our computed values were showing `sidecars` being set to {}, when it
should be [].

After tracing through our chart dependencies to find where
it's coming from, it turns out the older version of this exporter pulls
in an older version of the kafka chart, which by default will set
sidecars to {} rather than []

## Type of change

## How has this been tested?
I've deployed the chart to a test cluster and all was well 🙌 

## Checklist


hazzadous (changes requested):
Can we also address the TODO [here](https://github.com/PostHog/charts-clickhouse/blob/main/charts/posthog/values.yaml#L2046) as I believe the new version uses 1.4.2

hazzadous (approved):

PR #532, opened 2022-08-17
fix(nginx): set service-upstream to "true"

hazzadous:
Instead of having NGINX maintain it's own list of upstreams, we want it
instead to use the service ClusterIP such that k8s can handle e.g. cases
where pods are evicted.

Without this we end up with 5xx errors e.g. on deploys. See
https://github.com/kubernetes/ingress-nginx/issues/257 for more details.

## Type of change

## Checklist


ellie (approved):

guidoiaquinti:
Good 🔎  team (part #2). Good to confirm that's what we've suspected yesterday 💪

guidoiaquinti (approved):

PR #545, opened 2022-08-23
ci(k6s): remove flaky pluginEvent test

hazzadous:
To get to a point we can use tests as a merge condition, I'm disabling
this test.

## Type of change

## Checklist

### DIFF
--- a/ci/k6/ingestion-test.js
+++ b/ci/k6/ingestion-test.js
@@ -30,7 +30,7 @@ export let options = {
       executor: 'per-vu-iterations',
       vus: 1,           // Number of VUs to run concurrently.
       iterations: 1,    // only run a single iteration after generateEvents() completes
-      startTime: '220s',  // duration + gracefulStop of the above + 3m (to account for the ingestion buffer)
+      startTime: '40s',  // duration + gracefulStop of the above
     },
   },
   thresholds: {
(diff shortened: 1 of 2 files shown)


guidoiaquinti (approved):

guidoiaquinti:
cc @tiina303 @yakkomajuri @macobo can you suggest a better way to test e2e ingestion _post-buffer_?

macobo:
I'll take a look tomorrow but have you tried `CONVERSION_BUFFER_ENABLED` env var which would disable the conversion buffer?

hazzadous:
> I'll take a look tomorrow but have you tried CONVERSION_BUFFER_ENABLED env var which would disable the conversion buffer?

No, will give it a go separately from this PR.

guidoiaquinti:
> I'll take a look tomorrow but have you tried `CONVERSION_BUFFER_ENABLED` env var which would disable the conversion buffer?

We can try that but I think it could be nice to have an e2e test with buffer enabled (as it should be the default moving on?)

tiina303:
> We can try that but I think it could be nice to have an e2e test with buffer enabled (as it should be the default moving on?)

How about setting `BUFFER_CONVERSION_SECONDS = 1`

macobo:
I'd say setting conversion buffer for this test was incidental - we already have extensive test coverage for it in the core repo, if we wanted it here we should have a separate test rather than overload the main one.

Note that the conversion buffer applies to ~0.3% of events in the wild.

PR #550, opened 2022-08-24
Preview workflow to checkout repo and attempt helm deploy

ellie:
This will fail as I haven't set the auth yet, but let's see if we can at
least fail in the right way :D

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/pr-deploy.yaml
+++ b/.github/workflows/pr-deploy.yaml
@@ -24,13 +24,13 @@ jobs:
 
       - name: Checkout
         uses: actions/checkout@v2 # no way of getting the correct ref from the issue event, hence the below
-        if: steps.check.outputs.has-permission && ${{ github.event.comment.body == '@deploy'}} # looks like this needs to go on every step. meh.
+        if: steps.check.outputs.has-permission && ${{ github.event.comment.body == '!deploy'}} # looks like this needs to go on every step. meh.
 
       - name: Checkout Pull Request
         run: hub pr checkout ${{ github.event.issue.number }} # github cli is pre installed
         env:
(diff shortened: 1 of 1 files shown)


ellie:
I very much doubt GitHub Actions will run this without it being merged, but I can try right?

ellie:
@deploy

ellie:
!deploy

.github/workflows/pr-deploy.yaml
```diff
@@ -22,14 +22,34 @@ jobs:
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        if: steps.check.outputs.has-permission && ${{ github.event.comment.body == '!deploy'}}
+
+      - name: Declare variables that we can share across steps
+        id: vars
+        run: |
+          GIT_SHORT="$(git rev-parse --short HEAD)"
+          echo "::set-output name=git_short::${GIT_SHORT}"
+          echo "::set-output name=fqdn_record::${GIT_SHORT}.posthog.cc"
+        if: steps.check.outputs.has-permission && ${{ github.event.comment.body == '!deploy'}}
```
hazzadous, line 41: I think this should be the below but I'm not sure.

```suggestion
        if: ${{ steps.check.outputs.has-permission && github.event.comment.body == '!deploy' }}
```

ellie: I _think_ you might be right, kinda wishing I could test this without pushing it up 🙃

ellie:
!deploy

.github/workflows/pr-deploy.yaml
hazzadous: Yeah it's not a great dev experience :(

PR #558, opened 2022-08-25
Add PR preview deploys - just add the deploy label

ellie:
1) Deploy PRs to pr_num.pr.posthog.cc when the deploy label is added to
   a PR. Also deploy updates on push
2) Use the GitHub deploy API to nicely showcase this
3) Handle cleanup of the deployment when the PR is merged or closed

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/pr-deploy-cleanup.yaml
+++ b/.github/workflows/pr-deploy-cleanup.yaml
@@ -1,18 +1,9 @@
-# This workflow handles deploying a branch to the pull request cluster, for
-# testing purposes. It is locked down so that it can only be used by people who
-# are a member of the PostHog Github org
-
-name: PR - Preview Deploy
-
 on:
   pull_request:
-    types: 
-      - labeled
-      - synchronize
-
+    types: [closed]
 jobs:
-  deploy:
-    name: Deploy Preview
+  cleanup:
+    name: Cleanup merged deploy
     runs-on: ubuntu-latest
     steps:
       - name: Check user permission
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-deploy-cleanup.yaml
```diff
@@ -0,0 +1,63 @@
...
+  pull_request:
+    types: [closed]
+jobs:
+  cleanup:
+    name: Cleanup merged deploy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check user permission
+        id: check
+        uses: scherermichael-oss/action-has-permission@17f29510f1bf987b916c8cbb451566a56eed23f1
```
guidoiaquinti, line 11: Can we check if the user has permissions in some other way? Asking as we are passing a GitHub token to a GitHub Action that doesn't seems to be maintained (latest release is from 2020) and it's not from a well known maintainer (GitHub, Google, etc..)

ellie: Sure, can swap it out for some javascript I think! I did skim the code of the action + pinned to the SHA, but totally get the concern

.github/workflows/pr-deploy-cleanup.yaml
```diff
@@ -0,0 +1,78 @@
...
+                username: context.actor,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+              });
+
+              if (!resp || !resp.data || !resp.data.permission) {
+                member = false;
+              } else {
+                const permission = resp.data.permission;
+                member = permission == "admin" || permission == "write";
```
ellie, line 25: See: https://docs.github.com/en/rest/collaborators/collaborators#get-repository-permissions-for-a-user for the list of possible permissions

guidoiaquinti (approved):

PR #560, opened 2022-08-26
Explicitly delete chi

ellie:
Otherwise it doesn't seem to clean up very nicely. Also deactivate the
env before we wrap up deleting resources, so nobody tries to use a
half-deleted env.

## Type of change

## Checklist

### DIFF
--- a/.github/workflows/pr-deploy-cleanup.yaml
+++ b/.github/workflows/pr-deploy-cleanup.yaml
@@ -50,6 +50,15 @@ jobs:
         with:
           token: ${{ secrets.DIGITALOCEAN_ACCESS_TOKEN }}
 
+      - name: mark environment as deactivated
+        uses: bobheadxi/deployments@v1
+        with:
+          step: deactivate-env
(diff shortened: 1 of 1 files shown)


guidoiaquinti (approved):

hazzadous:
I suspect this is due to it not having the appropriate annotations for helm.sh @ellie . If the operator supports it we could add them in, which could be a nice more general solution.

hazzadous (approved):

hazzadous:
(but also, that can be for another day)

ellie:
> I suspect this is due to it not having the appropriate annotations for helm.sh @ellie . If the operator supports it we could add them in, which could be a nice more general solution.

I was hoping it would be cleaned up properly with this:

https://github.com/PostHog/charts-clickhouse/pull/560/files#diff-26113950b3f997a4c75c44dd07fd5fa767c0e3412731514a3e11fccd1547f56dR75

But apparently not :( 

Otherwise I totally agree, would be nice for it to "just work"

PR #567, opened 2022-09-01
chore: bump to version 1.39.1

pauldambra:
## Description
bumps posthog to version 1.39.1


guidoiaquinti (approved):

hazzadous (approved):
I believe you'll also have to add a redirect to posthog.com from 1.39.1 -> 1.39.0, unless you're going to add release notes here as well.

pauldambra:
> I believe you'll also have to add a redirect to posthog.com from 1.39.1 -> 1.39.0, unless you're going to add release notes here as well.

Oh! What release notes are they?

hazzadous:
@pauldambra I'm not sure if there are any, but previously I have added e.g. https://github.com/PostHog/posthog.com/blob/master/netlify.toml#L1415

Somewhere in the app there is a link that constructs a URL, and when we go to .1 release it breaks the link

PR #568, opened 2022-09-06
feat(ingress): allow setting ingress.hostname to empty

hazzadous:
Previously you would be able to set ingress.hostname to an empty string.
However, the SITE_URL would be set incorrectly, thus e.g. links to
sharable dashboards would no longer work.

We want to be able to set the hostname to empty such that the ingress
controller will pick up any traffic, no matter what the "Host" http
header. This was we can allow users to provide reverse proxies on
different addresses which are just proxying the original Host headers
through.

## Type of change

## Checklist


ellie (approved):

hazzadous:
Ooops, I broke _all_ the tests, looking.

PR #570, opened 2022-09-08
chore(loki): update to chart 3.0.1

hazzadous:
This is a big change and includes scaling components, via a distributed
storage backend e.g. S3.

I'm not sure how many people are using this so would like to make this a
minor patch rather than a major one, I think not many using, and at any
rate assuming no custom config has been used this should be ok if
introducing a little bit of downtime.

## Type of change

## Checklist


fuziontech (commented):
:partying_face:

PR #572, opened 2022-09-08
chore: reenable loki 3.x

hazzadous:
It seems that Grafana has created a new chart, grafana/loki 3.x, based on the simple scalable one from before. 
We were on the single binary version which was grafana/loki chart version 2.x, but to be able to scale better
we want to upgrade to 3.x.

To try to be a little backwards compat. we specify that we use filesystem storage, which should trigger using the single binary version. However, configuration will not be compatible if users have any custom values.

NOTE: Disable selfmonitor. It requires additional CRDs which aren't included on upgrade.

Tries to reinstate https://github.com/PostHog/charts-clickhouse/pull/570

Update notes [here](https://github.com/PostHog/posthog.com/pull/4196)

## Type of change

## Checklist


charts/posthog/values.yaml
```diff
@@ -1221,6 +1221,32 @@ grafana:
...
+  # this, although using the operator might offer a good option to allow
+  # decoupling of e.g. service and rule definitions from the Grafana and co
+  # workloads.
+  monitoring:
+    alerts:
+      enabled: false
+    rules:
+      enabled: false
+    selfMonitoring:
+      enabled: false
```
guidoiaquinti, line 1245:
```suggestion
      enabled: false # requires additional CRDs
```

guidoiaquinti (approved):

guidoiaquinti:
Upgrade fails 

```
Error: UPGRADE FAILED: cannot patch "posthog-loki" with kind StatefulSet: StatefulSet.apps "posthog-loki" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden
helm.go:84: [debug] cannot patch "posthog-loki" with kind StatefulSet: StatefulSet.apps "posthog-loki" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden
```

hazzadous:
> Upgrade fails

💣 

Will have to be a major upgrade I think, with a removal of the stateful set before upgrade. @guidoiaquinti do you have advice on how to handle? Is there a better approach?

guidoiaquinti:
Not really. My suggestion is to cut a major release with release notes indicating how to delete the statefulset.

Issue #410, opened 2022-06-01
clickhouse.enabled: false creates clickhouseinstall CRDs


gajus:
## Bug description

clickhouse.enabled: false creates clickhouseinstall CRDs

## Expected behavior

I would not expect CRDs to be created.

## Actual behavior

<img width="525" alt="Screen Shot 2022-06-01 at 6 50 34 PM" src="https://user-images.githubusercontent.com/973543/171519841-8e31620b-749e-4ad4-8147-50b5bf4fecb0.png">

## How to reproduce

1. use provided config

## Environment

- [ ] Deployment platform (AWS/Azure/DigitalOcean/GCP/...): _please provide_
- [ ] k8s version (server version from `kubectl version`): _please provide_
- [ ] Chart version/commit (`helm list -n posthog`): _please provide_
- [ ] Posthog version: _please provide_

guidoiaquinti:
👋 Hi again @gajus. This is true but it is unfortunately a limitation of Helm + how we package upstream resources from [Altinity/clickhouse-operator](https://github.com/Altinity/clickhouse-operator) as they don't offer (yet) a Helm chart (see https://github.com/Altinity/clickhouse-operator/pull/836, https://github.com/Altinity/clickhouse-operator/issues/837, https://github.com/Altinity/clickhouse-operator/issues/941).

If you want to skip the the CRD installation, my suggestion is to pass `--skip-crds` to your Helm command.

Here below is a better explanation of the problem from the [Helm official documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/):

> With the arrival of Helm 3, we removed the old `crd-install` hooks for a more simple methodology. There is now a special directory called `crds` that you can create in your chart to hold your CRDs. These CRDs are not templated, but will be installed by default when running a helm install for the chart. If the CRD already exists, it will be skipped with a warning. If you wish to skip the CRD installation step, you can pass the `--skip-crds` flag.
> 
> **Some caveats (and explanations)**
> There is no support at this time for upgrading or deleting CRDs using Helm. This was an explicit decision after much community discussion due to the danger for unintentional data loss. Furthermore, there is currently no community consensus around how to handle CRDs and their lifecycle. As this evolves, Helm will add support for those use cases.
> 
[...]

PR #573, opened 2022-09-08
fix: make clickhouse CRDs optional

romilpunetha:
## Description

When trying to add a self-hosted external clickhouse instance that is maintained by argocd, the CRDs generated by clickhouse operator are clashing with the ones generated by posthog charts. This change makes posthog clickhouse CRDs optional.

## Type of change

## How has this been tested?

1. Set `cloud=aws` and `clickhouse.installCRDs=false` in values.yaml
2. helm dep up
3. helm template . -f values.yaml
4. The output should not have clickhouse CRDs.

## Checklist

### DIFF
--- a/charts/posthog/templates/crds/clickhouseoperatorconfigurations.clickhouse.altinity.com.yaml
+++ b/charts/posthog/templates/crds/clickhouseoperatorconfigurations.clickhouse.altinity.com.yaml
@@ -2,6 +2,7 @@
 #
 # NONE
 #
+{{- if .Values.clickhouse.installCRDs }}
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
@@ -290,4 +291,4 @@ spec:
                         It can be set to a file and line number with a logging line.
                         Ex.: file.go:123
                         Each time when this line is being executed, a stack trace will be written to the Info log.
-
+{{- end }}
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -869,6 +869,8 @@ zookeeper:
 clickhouse:
   # -- Whether to install clickhouse. If false, `clickhouse.host` must be set
   enabled: true
+  # -- Whether to install `clickhouse` CRDs.
+  installCRDs: false
   # -- Which namespace to install clickhouse and the `clickhouse-operator` to (defaults to namespace chart is installed to)
   namespace:
   # -- Clickhouse cluster
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -870,7 +870,7 @@ clickhouse:
   # -- Whether to install clickhouse. If false, `clickhouse.host` must be set
   enabled: true
   # -- Whether to install `clickhouse` CRDs.
-  installCRDs: false
+  installCRDs: true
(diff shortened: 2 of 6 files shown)


charts/posthog/values.yaml
```diff
@@ -869,6 +869,8 @@ zookeeper:
 clickhouse:
   # -- Whether to install clickhouse. If false, `clickhouse.host` must be set
   enabled: true
+  # -- Whether to install `clickhouse` CRDs.
+  installCRDs: false
```
hazzadous, line 873: Worth changing this to `true` by default to avoid changing current behaviour, especially in a patch release.

fyi there are a few test [failures in CI](https://github.com/PostHog/charts-clickhouse/runs/8303362148?check_suite_focus=true), let us know if you need nay help running these tests locally.

charts/posthog/templates/crds/clickhouseoperatorconfigurations.clickhouse.altinity.com.yaml
```diff
@@ -2,6 +2,7 @@
 #
 # NONE
 #
+{{- if .Values.clickhouse.installCRDs }}
```
hazzadous, line 5: Looking at this change a little closer, I think this is a file that will be overwritten when we we run [`scripts/clickhouse_operator_sync.sh`](https://github.com/PostHog/charts-clickhouse/blob/main/scripts/clickhouse_operator_sync.sh). As such, it would be preferable to update that script over updating them here.

Have a look at the script [here](https://github.com/PostHog/charts-clickhouse/blob/main/scripts/clickhouse_operator_sync.sh#L61). Can you add something similar for the CRDs, perhaps just the addition of the crd files to `FILES`?

@guidoiaquinti what do you think

charts/posthog/values.yaml
```diff
@@ -869,6 +869,8 @@ zookeeper:
 clickhouse:
   # -- Whether to install clickhouse. If false, `clickhouse.host` must be set
   enabled: true
+  # -- Whether to install `clickhouse` CRDs.
+  installCRDs: true
```
guidoiaquinti, line 873: Not a huge fan of making this optional as we will likely end up supporting installations with CRDs not in sync with this chart.

romilpunetha:
Closing this PR. 
For setup via helm, the right way prevent CRD installation is to use `--skip-crds` flag. 
For argocd, the right way to do it is presented in this [link](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-skip-crds).

charts/posthog/values.yaml
romilpunetha: Not sure I fully understand this point. 
Wouldn't it be like `Posthog version x is compatible with clickhouse version >= y and <=z` 
How is it different from other helm charts that support services with CRDs, like istio?

guidoiaquinti: > Wouldn't it be like Posthog version x is compatible with clickhouse chart version >= y and <=z

That yes. I'm more worried about the scenario: 

- customer manually manages the CH operator CRDs
- customer upgrades the Helm chat without upgrading the CRDs if needed
- issues 

This issue is not present if we hard package everything together.

hazzadous:
For reference, this has been an issue previously as well: https://github.com/PostHog/charts-clickhouse/issues/410

Also, relevant issue in the clickhouse-operator repo: https://github.com/Altinity/clickhouse-operator/issues/941

PR #574, opened 2022-09-12
ci(monitoring): install loki/promtail/grafana in CI k3s deployments

hazzadous:
I was trying to [upgrade Loki Helm
chart](https://github.com/PostHog/charts-clickhouse/pull/570) to version
3.x (which is largely the previous loki-simple-scalable chart), which
would allow us to scale for more log bandwidth.

However, it requires manual installing of CRDs if we are upgrading.
Adding this to the k3s tests means we will also be testing the upgrade
and would have spotted this.

It does however mean that if there are any flakes caused by e.g.
additional memory requirements or another factor then it could be
annoying to have these e.g. block a release.

## Type of change

## Checklist

### DIFF
--- a/ci/values/k3s-plugin-server-split.yaml
+++ b/ci/values/k3s-plugin-server-split.yaml
@@ -36,3 +36,17 @@ plugins:
 # Enable pluginsAsync for this test
 pluginsAsync:
   enabled: true
+
+# Enable monitoring deployments. We are using these in our own deployments so
+# it's worth having these tested before they hit our clusters.
+loki:
+  enabled: true
+
+promtail:
+  enabled: true
+
+grafana:
+  enabled: true
+
+eventrouter:
+  enabled: true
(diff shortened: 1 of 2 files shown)


ellie (approved):

ci/values/k3s-plugin-server-split.yaml
```diff
@@ -36,3 +36,17 @@ plugins:
...
+  enabled: true
+
+promtail:
+  enabled: true
+
+grafana:
+  enabled: true
+
+eventrouter:
+  enabled: true
```
guidoiaquinti, line 52: Should we add prometheus and few exporters as well?

hazzadous: shoot, yes. I knew I was missing something, my mind was just on logs

hazzadous: Plus I was thinking that maybe I should enable just for our live version, if only to avoid e.g. having to be blocked by e.g. a dependency not supporting a specific version.

Will follow up as appropriate.

PR #575, opened 2022-09-12
dev(codespaces): get k3s working

hazzadous:
Instead of using containerd as the runtime, we use dockerd, otherwise I
was getting issues with using overlayfs. I tried to use
docker-from-docker as well, but that didn't seem to work due to not
being able to mount /etc/hosts.

## Type of change

## Checklist

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,11 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/ubuntu/.devcontainer/base.Dockerfile
+
+# [Choice] Ubuntu version (use ubuntu-22.04 or ubuntu-18.04 on local arm64/Apple Silicon): ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
+ARG VARIANT="jammy"
+FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
+
+# [Optional] Uncomment this section to install additional OS packages.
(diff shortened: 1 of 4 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/ubuntu/.devcontainer/base.Dockerfile
+
+# [Choice] Ubuntu version (use ubuntu-22.04 or ubuntu-18.04 on local arm64/Apple Silicon): ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
+ARG VARIANT="jammy"
```
guidoiaquinti, line 4: Good name

.devcontainer/setup_container.sh
```diff
@@ -3,9 +3,9 @@
 set -e -o pipefail
 
 # Install k3s
-curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=v1.22 INSTALL_K3S_EXEC="--disable=traefik --disable=metrics-server" sh -
+curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=v1.22 INSTALL_K3S_SKIP_ENABLE=true INSTALL_K3S_SKIP_ENABLE=true sh -
 sudo apt-get update
```
guidoiaquinti, line 7: You could add as well `DEBIAN_FRONTEND=noninteractive` to all `apt` invocations

PR #594, opened 2022-10-12
Added additional backend for static web files

davordragic:
## Description
When trying to hide Posthog web interface behind Google IAP (GCP) you have issues with static files which cannot be served.
Idea was to create additional backend and route path, so static files can still be public, while Posthog login is hidden.

## Type of change

## How has this been tested?
Tested in our test and production enviroment.

## Checklist

### DIFF
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -63,6 +63,20 @@ spec:
                 name: {{ template "posthog.fullname" . }}-web
                 port:
                   number: {{ .Values.service.externalPort }}
+          {{- if .Values.ingress.separateStatic -}}  
+          {{- if (ne (include "ingress.type" .) "clb") }}
+          - pathType: Prefix
+            path: "/static"
+          {{- else }}
+          - pathType: ImplementationSpecific
+            path: "/static/*"
+          {{- end }}
+            backend:
+              service:
+                name: {{ template "posthog.fullname" . }}-web-static
+                port:
+                  number: {{ .Values.service.externalPort }}
+          {{- end -}}
           {{- if (ne (include "ingress.type" .) "clb") }}
           - pathType: Prefix
             path: "/batch"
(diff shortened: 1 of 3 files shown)


guidoiaquinti:
👋  Hey! Sorry for the late reply here. If you want to separate static assets serving from dynamic requests I suggest you to leverage the PostHog `JS_URL` environment variable instead of creating a new service and load balancer route specific to it. 

Also, your implementation is for GCP only while this chart needs to supports AWS, Azure, DigitalOcean and Google Cloud Platform.

davordragic:
Hej! Thank you for replay.
Fair point about JS_URL environment variable, will use that.

But meanwhile, notice that we have (moved from cloud) events going to host.url/events/ and are routed to web service... At the moment, I used same aproach to send /engage/ to my "static-web" which is not behind IAP.

Is there any other idea, how to move /engage/ like static url?

guidoiaquinti:
I'm not sure if I understand what you want to archive. Do you want to move specific endpoints outside static assets?

davordragic:
A bit of context would help.

At the moment I have LB (Load balancer) and IAP (Identity-Aware Proxy) over web service.
image

Because web service is behind IAP, /engage/ and /static/ are not accessable for third party users and we are not able to get any data from them. (xhr reqest gets 401 http reposnse).
image

My idea was to create "static-web" service and not put it behind IAP ^ lika I did.
But since there is better solution for JS_URL, it's cleaner solution (and better). 

With same idea, would be possible to route /engage/ url path to another service, so it does not hit IAP and 401 response?
PS: I have no idea, how we got url path with /engage/ -> is it possible just to replace this with any of public api (service: events)

guidoiaquinti:
> With same idea, would be possible to route /engage/ url path to another service, so it does not hit IAP and 401 response?
PS: I have no idea, how we got url path with /engage/ -> is it possible just to replace this with any of public api (service: events)

That's a bug, it should be fixed by https://github.com/PostHog/charts-clickhouse/pull/602

davordragic:
Great. Than issues solved. Just need to see how JS_URL works and that's it.

This PR can be closed.

PR #598, opened 2022-10-18
refactor(plugins-server): templatize plugin server deployment

hazzadous:
Updates the chart to:

1. include pluginsExports, pluginsJob, pluginsScheduler, pluginsIngestion deployments which can be used instead of the plugins and pluginsAsync deployments.
2. the existing deployments should work as before, e.g. the plugins and pluginsAsync deployments
3. you can enable these deployments with: 
```
plugins:
    enabled: false

pluginsExports:
    enabled: true

...
```
You should receive an error if you enable both the plugins/pluginsAsync and the new ones at the same time.

There is some awkwardness in that the tests run against the release version of posthog, so this will only work once a release has happened, hence in the k3s tests we still only test the legacy functionality. Production tracks the latest release however, so we will be able to enable it there.

The corresponding updates to the app are at https://github.com/PostHog/posthog/pull/12191

## Type of change

## Checklist


yakkomajuri (approved):
"You should receive an error if you enable both the plugins/pluginsAsync and the new ones at the same time." --> maybe we actually want to do this at first to account for potential issues when flipping over?

Otherwise, I mean, _looks_ good to _me_ - stuff seems sensible, but maybe a second set of eyes could guarantee nothing will implode from this?

guidoiaquinti:
> There is some awkwardness in that the tests run against the release version of posthog, so this will only work once a release has happened, hence in the k3s tests we still only test the legacy functionality. Production tracks the latest release however, so we will be able to enable it there.

LGTM, can we anyway add a kubetest pointing to `image.tag: latest` once https://github.com/PostHog/posthog/pull/12191 is merged? This will help us to temporary test this scenario, after the release is out we can remove the custom override from it

hazzadous:
> LGTM, can we anyway add a kubetest pointing to image.tag: latest once https://github.com/PostHog/posthog/pull/12191 is merged? This will help us to temporary test this scenario, after the release is out we can remove the custom override from it.

@guidoiaquinti Have added this, do you think this is enough? https://github.com/PostHog/charts-clickhouse/pull/598/files#diff-ee31744c16c09322dc783a319da3c759dd38bf18f444b8f5f1ed0fb4ef49e65cR52

guidoiaquinti:
> > LGTM, can we anyway add a kubetest pointing to image.tag: latest once [PostHog/posthog#12191](https://github.com/PostHog/posthog/pull/12191) is merged? This will help us to temporary test this scenario, after the release is out we can remove the custom override from it.
> 
> @guidoiaquinti Have added this, do you think this is enough? https://github.com/PostHog/charts-clickhouse/pull/598/files#diff-ee31744c16c09322dc783a319da3c759dd38bf18f444b8f5f1ed0fb4ef49e65cR52

Thank you!

hazzadous:
> maybe we actually want to do this at first to account for potential issues when flipping over?

@yakkomajuri I have applied this to dev so we have somewhere we can validate that it's functional before [deploying to prod](https://github.com/PostHog/posthog-cloud-infra/pull/684). Which bits and how can we test that it's all working?

PR #602, opened 2022-10-20
Ingress: move `/engage` calls from web to events

guidoiaquinti:
## Description
While reviewing [this](https://github.com/PostHog/charts-clickhouse/pull/594#issuecomment-1285608933) customer report I've found that we are incorrectly routing `/engage` calls to the default deployment (`posthog-web`) instead of `posthog-events`.

## Type of change

## How has this been tested?
CI

## Checklist

### DIFF
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -86,6 +86,9 @@ spec:
             path: "/e"
             backend: *INGESTION
           - pathType: Prefix
+            path: "/engage"
+            backend: *INGESTION
+          - pathType: Prefix
             path: "/track"
             backend: *INGESTION
           - pathType: Prefix
@@ -100,6 +103,9 @@ spec:
             backend: *INGESTION
           - pathType: ImplementationSpecific
             path: "/e/*"
+            backend: *INGESTION
+          - pathType: ImplementationSpecific
+            path: "/engage/*"
             backend: *INGESTION
           - pathType: ImplementationSpecific
             path: "/track/*"
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -11,7 +11,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 27.2.0
+version: 27.2.1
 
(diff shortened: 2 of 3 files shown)


charts/posthog/templates/ingress.yaml
```diff
@@ -85,6 +85,9 @@ spec:
           - pathType: Prefix
             path: "/e"
```
hazzadous, line 86: Maybe do this instead?

```suggestion
            path: "/e/"
```

Otherwise we have e.g. /events/ going to posthog-events as well which isn't correct.

hazzadous: (although I guess we need to make sure that you need the trailing slash for that endpoint 🤔 )

hazzadous (approved):
LGTM as resolves the immediate issue

charts/posthog/templates/ingress.yaml
guidoiaquinti: I agree but that's fundamentally another routing issue 😁 

Should we change all the non CLB with

```
diff --git a/charts/posthog/templates/ingress.yaml b/charts/posthog/templates/ingress.yaml
index ca04d8b..ad06f99 100644
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -77,22 +77,22 @@ spec:
                   number: {{ .Values.service.externalPort }}
           - pathType: Prefix
-            path: "/capture"
+            path: "/capture/"
             backend: *INGESTION
           - pathType: Prefix
-            path: "/decide"
+            path: "/decide/"
             backend: *INGESTION
           - pathType: Prefix
-            path: "/e"
+            path: "/e/"
             backend: *INGESTION
           - pathType: Prefix
-            path: "/engage"
+            path: "/engage/"
             backend: *INGESTION
           - pathType: Prefix
-            path: "/track"
+            path: "/track/"
             backend: *INGESTION
           - pathType: Prefix
-            path: "/s"
+            path: "/s/"
             backend: *INGESTION
           {{- else }}
           - pathType: ImplementationSpecific
```

?

guidoiaquinti: I'm inclined to merge this as it is and re-evaluate the issue with the other endpoints. Fair point tho!

PR #604, opened 2022-10-21
CI - k6 ingestion test: use nginx

guidoiaquinti:
## Description
While working on another task, I realised that for the CI k6 ingestion test we install the ingress and ingress-nginx despite we don't use them in CI.

This PR points the CI k6 test to run against the ingress instead of directly hitting the pods (so that we can also test the routing to them). 

## Type of change

## How has this been tested?
CI

## Checklist

### DIFF
--- a/ci/values/k3s-plugin-server-ingestion-async-split.yaml
+++ b/ci/values/k3s-plugin-server-ingestion-async-split.yaml
@@ -1,9 +1,10 @@
 cloud: k3s
 
 image:
-  tag: latest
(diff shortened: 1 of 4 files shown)


ci/values/k3s-plugin-server-ingestion-async-split.yaml
```diff
@@ -1,7 +1,7 @@
 cloud: k3s
 
 image:
-  tag: latest
+  tag: latest # TODO: remove this override once 1.41.0 is the default
```
hazzadous, line 4: 👍

PR #610, opened 2022-10-31
chore(posthog): release posthog 1.41.0

hazzadous:
We're releasing 1.41.0 and as such we need to update the default image
in the chart.

NOTE: remember to remove `-unstable` from the image tag in values.yaml
before merging this.

## Type of change
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #612, opened 2022-11-02
chore: Bump version to 1.41.1

mariusandra:
## Description

Update chart for 1.41.0

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -15,7 +15,7 @@ version: 29.0.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 1.41.0
+appVersion: 1.41.1
 
 dependencies:
   - name: cert-manager
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -11,11 +11,11 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 28.0.0
+version: 29.0.0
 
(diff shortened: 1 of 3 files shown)


mariusandra:
After making this PR I also see https://github.com/PostHog/charts-clickhouse/pull/610 - not sure what should get merged.
Here I bumped the major version based on @tiina303 's [comment here](https://posthog.slack.com/archives/C0374DA782U/p1667316396423159?thread_ts=1667315034.067579&cid=C0374DA782U)

charts/posthog/Chart.yaml
```diff
@@ -11,11 +11,11 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 28.0.0
```
guidoiaquinti, line 14: I don't think we need to bump the major version if we don't have release notes and/or breaking changes

mariusandra: It's coming here:
https://github.com/PostHog/posthog.com/pull/4562

mariusandra: The "here" will link to the array "[here](https://github.com/PostHog/posthog.com/blob/f153fd92fda5dfb26617db44e40fc6ec43ef3c5b/contents/blog/the-posthog-array-1-41-0.md#1410-update-guide-for-self-hosted-users)"

ellie (approved):

charts/posthog/Chart.yaml
tiina303: https://github.com/PostHog/posthog.com/pull/4562 WIP here

PR #569, opened 2022-09-07
Add support for Kubernetes v1.25, drop v1.22

guidoiaquinti:
## Description
This PR adds the support for Kubernetes v1.25 and drops the support for v1.22 as [EOL since 2022-10-28](https://kubernetes.io/releases/#release-v1-22).

Accompanying release notes are available at: https://github.com/PostHog/posthog.com/pull/4513

## Type of change
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## How has this been tested?
CI is ✅ 

## Checklist
- [ ] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #615, opened 2022-11-04
Allow deployment on kube 1.22

georgekaz:
Google still run GKE v1.22. Please be considerate and allow Posthog to continue to be deployed on 1.22

https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -11,7 +11,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 29.0.0
+version: 29.0.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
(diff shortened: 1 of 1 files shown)


guidoiaquinti:
👋 Hi @georgekaz and thank you for your contribution! 

I’ve recently replied to a related comment [here](https://github.com/PostHog/charts-clickhouse/pull/569#issuecomment-1302242327), please let me know if it makes sense to you as well. 

v1.23 is available in GCP in both rapid (early adopters) and regular (early majority) [release channels](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels#what_channel_should_i_use). If you only want to run on stable channel (understandable), at the bottom of the comment above you can find an easy workaround for your issue.

georgekaz:
Hi @guidoiaquinti . Although I appreciate that you are a small team and have limited versions you can support, taking into consideration that none of GKE, EKS or AKS support 1.25, but all of them support 1.22, why have you chosen to support 1.25 rather than 1.22? In my opinion it's a bad choice.

We could use the workaround with little problem, although I'd more likely just update the image tag to the latest releases and leave the helm chart where it is. But I still feel like this is a change people shouldn't be forced to make.

guidoiaquinti:
> why have you chosen to support 1.25 rather than 1.22? In my opinion it's a bad choice.

The reason is that we ship a Helm chart for the Kubernetes platform and as you can see [here](https://kubernetes.io/releases/) the current supported versions are 1.23, 1.24 and 1.25.

> We could use the workaround with little problem, although I'd more likely just update the image tag to the latest releases and leave the helm chart where it is. But I still feel like this is a change people shouldn't be forced to make.

Yes, that's another valid workaround.

georgekaz:
I know that 1.25 is there but it's only 3 weeks old, and there are way more people using 1.22 than using 1.25. 1.22 maybe EOL in the home-rolled version, but on the cloud platforms it's still supported. I really think you should discuss this internally and reconsider. You're basically making it unnecessarily harder for anyone using one of the major cloud platform's k8s offerings.

PR #622, opened 2022-11-11
Add the ability to configure access_management capabilities

fuziontech:
## Description
We have a paying user requesting to have the ability to create users dynamically using SQL
https://clickhouse.com/docs/en/sql-reference/statements/create/user/

In order to do this we must enable `access_management` in the `users.xml` configs.
https://clickhouse.com/docs/en/operations/settings/settings-users#access_management-user-setting

This PR makes it possible to enable this through the following values:
```yaml
clickhouse:
  enable: true
  additionalUsersConfig:
    admin/access_management: "1"
```

Requested by customer here:
https://posthog.slack.com/archives/C03G21222QY/p1668109138746399?thread_ts=1666628600.482729&cid=C03G21222QY

## Type of change

## How has this been tested?
Tested on a Digitalocean K8s cluster 

## Checklist

### DIFF
--- a/charts/posthog/templates/clickhouse_instance.yaml
+++ b/charts/posthog/templates/clickhouse_instance.yaml
@@ -17,6 +17,7 @@ spec:
       {{ .Values.clickhouse.backup.backup_user }}/networks/ip: "0.0.0.0/0"
       {{ template "clickhouse.backupPasswordValue" . }}
       {{- end}}
+      {{- merge dict .Values.clickhouse.users .Values.clickhouse.defaultUsers | toYaml | nindent 6 }}
     profiles:
       {{- merge dict .Values.clickhouse.profiles .Values.clickhouse.defaultProfiles | toYaml | nindent 6 }}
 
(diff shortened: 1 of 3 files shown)


charts/posthog/templates/clickhouse_instance.yaml
```diff
@@ -17,6 +17,9 @@ spec:
       {{ .Values.clickhouse.backup.backup_user }}/networks/ip: "0.0.0.0/0"
       {{ template "clickhouse.backupPasswordValue" . }}
       {{- end}}
+      {{- if .Values.clickhouse.users }}
```
guidoiaquinti, line 20: I find this a bit confusing as we have already other values (like: `clickhouse.user` defined). 

What about `additionalUsersConfig`?

charts/posthog/values.yaml
```diff
@@ -39,12 +39,10 @@ env: []
 #   - name: FOO
 #     value: bar
 
-
```
guidoiaquinti, line 42: Lots of formatting changes were committed unrelated to this change? Not sure if this was expected.

fuziontech: It was. My IDE automatically formats yaml. Seems like most of this makes sense to me.

hazzadous (approved):

hazzadous:
Have made a format updates as well as including the new value in the values.yaml cc @fuziontech @guidoiaquinti 😓

guidoiaquinti (approved):

PR #630, opened 2022-11-16
Add imagePullSecrets

guidoiaquinti:
## Type of change

## Checklist

### DIFF
--- a/charts/posthog/grafana-dashboards/pgbouncer.json
+++ b/charts/posthog/grafana-dashboards/pgbouncer.json
@@ -24,7 +24,7 @@
                     "type": "loki",
                     "uid": "P8E80F9AEF21F6940"
                 },
-                "enable": true,
+                "enable": false,
                 "expr": "{namespace=\"posthog\", container=~\"grafana-annotation-job-.*\"}",
                 "hide": false,
                 "iconColor": "dark-blue",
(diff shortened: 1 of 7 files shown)


charts/posthog/templates/worker-deployment.yaml
```diff
@@ -33,31 +33,41 @@ spec:
         {{- end }}
     spec:
       serviceAccountName: {{ template "posthog.serviceAccountName" . }}
+
       {{- if .Values.worker.affinity }}
       affinity:
 {{ toYaml .Values.worker.affinity | indent 8 }}
       {{- end }}
+
```
ellie, line 41: Formatting changes because otherwise this is a giant block of "omg I cannot read it"

charts/posthog/templates/_snippet-plugins-deployment.tpl
```diff
@@ -37,31 +37,47 @@ spec:
...
       priorityClassName: "{{ .params.priorityClassName }}"
       {{- end }}
+
       {{- if .root.Values.image.imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml .root.Values.image.imagePullSecrets | nindent 8 }}
       {{- end }}
+
+      # I do not know for sure if the old one has been used anywhere, so do both :(
+      {{- if .root.Values.image.pullSecrets }}
```
hazzadous, line 70: Should this be an if else with the above definition?

What's the `imagePullSecrets` / `pullSecrets` distinction?

ellie: `imagePullSecrets` is something that's been here for ages, but not documented and idk if it has been used.

`pullSecrets` is what has been added now - `image.pullSecrets` is imo nicer than `image.imagePullSecrets`, and naming is inline with our other `image.*` things. I'd like to deprecate/remove `imagePullSecrets` eventually

It could probably be an else, however

hazzadous (approved):
Looks good. Maybe should add tests for the plugin-server workloads as well however?

ellie:
> Maybe should add tests for the plugin-server workloads as well however?

Agreed, I'll do this now + bump the chart version! Thanks for the review 🙌

PR #632, opened 2022-11-18
Replace ClusterRole with Role only for posthog namespace

SuperTux88:
## Description

According to https://github.com/Altinity/clickhouse-operator/issues/994#issuecomment-1227132257 it isn't required to give clickhouse-operator permissions for the whole cluster if the clickhouse-operator is only deployed into a single namespace, which is the case for PostHog. So it's better for security reasons to only give the permissions which are required and not just to everything.

## Type of change

## How has this been tested?

Unit tests and installation in our cluster.

## Checklist


guidoiaquinti:
👋 @SuperTux88 and thank you for your contribution! We currently vendor everything in `charts/posthog/templates/clickhouse-operator` via [scripts/clickhouse_operator_sync.sh](https://github.com/PostHog/charts-clickhouse/blob/main/scripts/clickhouse_operator_sync.sh) as there isn't an Helm chart for the `clickhouse-operator`. Due to this we usually highly discourage to directly edit templates in that folder.

SuperTux88:
Thanks for your comment, I didn't know about this script. I also updated the script now to use `MANIFEST_PRINT_RBAC_NAMESPACED` to generate the roles, so there are no manual changes required for the templates in that folder.

scripts/clickhouse_operator_sync.sh
```diff
@@ -18,13 +18,13 @@ TMP_FOLDER="$(mktemp -d)"
 trap 'rm -rf -- "$TMP_FOLDER"' EXIT
 
 CLICKHOUSE_OPERATOR_TAG="0.18.4"
-URL="https://raw.githubusercontent.com/Altinity/clickhouse-operator/${CLICKHOUSE_OPERATOR_TAG}/deploy/operator/clickhouse-operator-install-template.yaml"
+REPO_URL="https://github.com/Altinity/clickhouse-operator.git"
```
guidoiaquinti, line 21: Why is this needed?

SuperTux88: Because running `cat-clickhouse-operator-install-yaml.sh` requires multiple template files, so instead of downloading all of them separately, it's easier to just clone the repo and have everything that is required there.

PR #641, opened 2022-11-24
Disable internal_metrics by default

macobo:
https://github.com/PostHog/posthog/pull/12943 removes the feature completely and it hasnt been useful for months. For old releases, keep the environment variable around


ellie (changes requested):
could we also bump the chart please? 🙏

ellie (approved):

PR #642, opened 2022-11-28
PgBouncer: use a real FQDN

guidoiaquinti:
## Description
In the PR that introduced the PgBouncer service FQDN we missed a `.` at the end to make it... an actual FQDN.

## Type of change

## How has this been tested?
CI is ✅ 

## Checklist


ellie (approved):
This _looks_ good, but after last week let's be super cautious rolling it out 😂

guidoiaquinti:
> This _looks_ good, but after last week let's be super cautious rolling it out 😂

Anything in particular we should take a look? The fqdn is only used as part of the init container (for now)

PR #644, opened 2022-11-29
chore: set Strict-Transport-Security header

danielxnj:
## Description
The security header Strict-Transport-Security is disabled; we need to enable it to force the site to use HTTPS.
This change sets the header with a max-age of 30 seconds.

## Type of change

## How has this been tested?
In a self-hosted installation, we can see the header is set:

image

## Checklist

### DIFF
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -19,6 +19,8 @@ metadata:
     {{- end }}
    {{- end }}
    {{- if (and (eq (include "ingress.type" .) "nginx") .Values.ingress.nginx.redirectToTLS "true") }}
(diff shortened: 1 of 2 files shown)


guidoiaquinti:
LGTM but I'm seeing the PR description related to the test blank. Can we make sure we test this on a self-hosted installation before merging?

danielxnj:
> LGTM but I'm seeing the PR description related to the test blank. Can we make sure we test this on a self-hosted installation before merging?

I have updated the description with the test I did in a self-hosted instance. I did 30 seconds of max Age, but to pass the check, we need to have 30 days min.

image

I suggest we keep this for a few days and them bump it to 30 days.

ellie (approved):

PR #648, opened 2022-12-13
Allow setting envvars in the events deployment

xvello:
## Description

Allow to set envvars to the events deployment, separate from the web deployment. This will be used to tune this deployment's behavior and performance.

Because this deploy already uses `.Values.web.env`, the new `.Values.events.env` list is added on top of it for backwards-compatibility. I am updating `.Values.web.env`'s doc to reflect that it also impacts events.

## Type of change

## How has this been tested?

Introduced a new unit test

## Checklist

### DIFF
--- a/charts/posthog/templates/events-deployment.yaml
+++ b/charts/posthog/templates/events-deployment.yaml
@@ -116,6 +116,9 @@ spec:
 {{- if .Values.web.env }}
 {{ toYaml .Values.web.env | indent 8 }}
 {{- end }}
+{{- if .Values.events.env }}
+{{ toYaml .Values.events.env | indent 8 }}
+{{- end }}
 
         {{- include "snippet.web-deployments.lifecycle" . | nindent 8 }}
 
--- a/charts/posthog/tests/events-deployment.yaml
+++ b/charts/posthog/tests/events-deployment.yaml
@@ -183,6 +183,32 @@ tests:
             name: SENTRY_DSN
             value: sentry.endpoint
 
+  - it: sets env vars from global, web and events sections
+    template: templates/events-deployment.yaml # TODO: remove once secrets.yaml will be fixed/removed
+    set:
+      cloud: local # TODO: remove once secrets.yaml will be fixed/removed
+      env:
+            - name: FIRST
+              value: one
+      web:
+        env:
+            - name: SECOND
+              value: two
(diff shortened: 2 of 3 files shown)


charts/posthog/templates/events-deployment.yaml
```diff
@@ -115,6 +115,9 @@ spec:
 {{- end }}
 {{- if .Values.web.env }}
 {{ toYaml .Values.web.env | indent 8 }}
+{{- end }}
```
guidoiaquinti, line 118: In order to avoid key collisions, what about having `.Values.events.env` first and then fallback with an elsif `.Values.web.env` to the current default?

charts/posthog/tests/events-deployment.yaml
```diff
@@ -183,6 +183,32 @@ tests:
             name: SENTRY_DSN
             value: sentry.endpoint
 
+  - it: sets env vars from global, web and events sections
```
guidoiaquinti, line 186: Can we add a test with both `web.env` and `events.env`? I think with the current logic we will hit a key collision if we specify the same key in both places (I might be wrong tho!)

charts/posthog/tests/events-deployment.yaml
```diff
@@ -183,6 +183,61 @@ tests:
             name: SENTRY_DSN
             value: sentry.endpoint
 
+  - it: uses web.env is events.env is empty
```
guidoiaquinti, line 186:
```suggestion
  - it: uses web.env if events.env is empty
```

charts/posthog/tests/events-deployment.yaml
```diff
@@ -183,6 +183,61 @@ tests:
...
+          content:
+            name: FIRST
+            value: one
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: SECOND
+            value: two
+
+  - it: ignores web.env is events.env is set
```
guidoiaquinti, line 209:
```suggestion
  - it: ignores web.env if events.env is set
```

guidoiaquinti (approved):
LGTM. Note that you need to bump the chart version in order to release a new version.

guidoiaquinti (approved):

hazzadous (approved):

PR #651, opened 2022-12-14
feat(alerts): add alert on exports consumer group backlog

hazzadous:
Adds a critical alert that checks to see it the delta in consumer lag
between now and 10 minutes ago is consistently >0 for 10 minutes
straight.

There are possibly cases where this doesn't work. I did the detla as I
didn't want to add an arbitrary message backlog value that might not
translate to all deployments of the app.

This should however catch the case were the lag is strictly
monotonically increasing.

I've also added lag alerts for:

 1. initial events ingestion
 2. clickhouse ingestion for events, persons, groups, person_distinct_id

This is in relation to https://posthog.slack.com/archives/C04F42Y4JRJ/p1670954181738359

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -1774,6 +1774,60 @@ prometheus:
                 summary: Exports and WebHooks backlog appears to have been increasing for 10 minutes straight.
                 description: "The backlog on consumer group clickhouse-plugin-server-async for topic clickhouse_events_json is increasing."
 
+            - alert: ClickHouseEventsConsumerLagIncreasing
+              expr: sum(delta(kafka_consumergroup_lag{topic="clickhouse_events_json", consumergroup="group1"}[10m])) > 0
+              for: 10m
+              labels:
+                severity: critical
+              annotations:
+                summary: ClickHouse events ingestion lag has been increasing for 10 minutes straight.
+                description: "The backlog on consumer group group1 for topic clickhouse_events_json is increasing."
+
+            - alert: ClickHousePersonConsumerLagIncreasing
+              expr: sum(delta(kafka_consumergroup_lag{topic="clickhouse_person", consumergroup="group1"}[10m])) > 0
+              for: 10m
+              labels:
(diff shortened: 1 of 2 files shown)


xvello (approved):

charts/posthog/values.yaml
```diff
@@ -1763,6 +1763,71 @@ prometheus:
...
+            - alert: ClickHousePersonConsumerLagIncreasing
+              expr: sum(delta(kafka_consumergroup_lag{topic="clickhouse_person", consumergroup="group1"}[10m])) > 0
+              for: 10m
+              labels:
+                severity: critical
+              annotations:
+                summary: ClickHouse person ingestion lag has been increasing for 10 minutes straight.
+                description: "The backlog on consumer group group1 for topic clickhouse_person is increasing."
+
+            - alert: ClickHousePersonConsumerLagIncreasing
```
xvello, line 1795:
```suggestion
            - alert: ClickHouseGroupsConsumerLagIncreasing
```

Copy-pasta error, due to too many identical alerts ^^ Maybe merge the 4 `group1` ones, with `sum(delta(kafka_consumergroup_lag{topic=~"clickhouse_events_json|clickhouse_person|clickhouse_groups|clickhouse_person_distinct_id", consumergroup="group1"}[10m])) by (topic)`, if we already have monitors that alert on different contexts

hazzadous: Yeah I was keeping separate just in case it was useful to set more descriptive summaries/descriptions, but I think it doesn't.

tiina303:
can we add runbook links, the runbook could just be very bare-bones initially, but for this we already know that finding a bad actor could help - both webhooks or exports

hazzadous:
> can we add runbook links, the runbook could just be very bare-bones initially, but for this we already know that finding a bad actor could help - both webhooks or exports

Yes 💯  would you suggest these be public runbooks? Perhaps we can add a GitHub Pages to this repo.

PR #656, opened 2022-12-19
fix(ingress): ensure only capture and decide go to events deployment

hazzadous:
Previously we would match e.g. any path that started with `/e` and route
this to the `events` pods. This isn't desired as this captures e.g.
`/events/` which is a user facing UI. Likewise for the other endpoints
that were previously k8s ingress `Prefix` rules.

Note that the ingress rules only match on path i.e. before any query
string components and the `?` character.

NOTE: I believe we do _not_ have any data that is passed in the request path
so I think we are safe to do an Exact.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -78,24 +78,54 @@ spec:
                 port:
                   number: {{ .Values.service.externalPort }}
           {{- if (ne (include "ingress.type" .) "clb") }}
+          # Match any url with a prefix ending with a forward-slash. Previously
+          # we would match without the forward-slash. To ensure that we pick up
+          # with the forward-slash omitted we also match Exact. Note tat we
+          # could use e.g. [ingress-nginx regex
+          # support](https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/#regular-expression-support)
+          # but to keep this agnostic we use a combination of Prefix and Exact.
           - pathType: Prefix
+            path: "/capture/"
+            backend: *INGESTION
+          - pathType: Exact
             path: "/capture"
             backend: *INGESTION
+
           - pathType: Prefix
+            path: "/decide/"
+            backend: *INGESTION
+          - pathType: Exact
             path: "/decide"
             backend: *INGESTION
+
           - pathType: Prefix
+            path: "/e/"
+            backend: *INGESTION
+          - pathType: Exact
             path: "/e"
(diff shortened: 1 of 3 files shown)


charts/posthog/templates/ingress.yaml
```diff
@@ -78,24 +78,54 @@ spec:
                 port:
                   number: {{ .Values.service.externalPort }}
           {{- if (ne (include "ingress.type" .) "clb") }}
+          # Match any url with a prefix ending with a forward-slash. Previously
+          # we would match without the forward-slash. To ensure that we pick up
+          # with the forward-slash omitted we also match Exact. Note tat we
```
fuziontech:
```suggestion
          # with the forward-slash omitted we also match Exact. Note that we
```

charts/posthog/templates/ingress.yaml
```diff
@@ -78,24 +78,54 @@ spec:
...
+          # could use e.g. [ingress-nginx regex
+          # support](https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/#regular-expression-support)
```
fuziontech, line 85:
```suggestion
          # could use e.g. [ingress-nginx regex support](https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/#regular-expression-support)
```

charts/posthog/templates/ingress.yaml
```diff
@@ -78,24 +78,54 @@ spec:
...
             path: "/track"
             backend: *INGESTION
+
           - pathType: Prefix
+            path: "/s/"
+            backend: *INGESTION
+          - pathType: Exact
             path: "/s"
             backend: *INGESTION
+
```
fuziontech, line 132: from @xvello: we should add `/i/` preemptively so that we can slowly move these endpoints under one scope and avoid an explosion of special endpoint routing cases here.

fuziontech (approved):
a few fixes and comments in here + a suggestion to add `/i/` so that we can avoid having an explosion in special case prefixes here.

charts/posthog/templates/ingress.yaml
hazzadous: I'll leave that for separate implementation for when it's implemented.

hazzadous:
Looks like there is a test failing re. kafka chart version not being available. That is unrelated to this change so will push up a PR separately for this.

PR #666, opened 2023-01-11
chore(session-recordings): separate session recordings from events

hazzadous:
We currently have session recordings events going to the same endpoint
as analytics events and decide endpiont. The characteristics of session
recordings is quite different, and our expectations of priority. It's
much more important that we capture events and service decide endpoints
than we capture all session recordings.

This separates the deployment to offer more flexibility in providing
different service availability.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -105,6 +105,9 @@ sessionRecordings:
     # for configuration options
     behavior:
 
+  # -- Resource limits for session recordings service.
+  resources: {}
+
   # -- Additional env variables to inject into the session recordings stack, uses `web.env` if empty.
   env: []
 
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -84,6 +84,37 @@ events:
   podSecurityContext:
     enabled: false
 
+sessionRecordings:
+  # -- Whether to install the PostHog events stack or not.
+  enabled: true
+
+  # -- Count of events pods to run. This setting is ignored if `events.hpa.enabled` is set to `true`.
+  replicacount: 1
+
+  hpa:
+    # -- Whether to create a HorizontalPodAutoscaler for the events stack.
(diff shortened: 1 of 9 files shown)


tomasfarias (approved):

charts/posthog/values.yaml
```diff
@@ -84,6 +84,40 @@ events:
...
+    # -- Set the HPA behavior. See
+    # https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/
+    # for configuration options
+    behavior:
+
+  # -- Resource limits for session recordings service.
+  resources: {}
+
+  # -- Additional env variables to inject into the session recordings stack, uses `web.env` if empty.
+  env: []
```
pauldambra, line 112:
```suggestion
  env: [
      DEPLOYMENT_ROLE: "SESSION_RECORDINGS_CAPTURE"
  ]
```

pauldambra:
I came poking about in this repo to figure out how to make exactly this and this PR is sat here like it's already xmas 🎉 

I'm guessing it's safe to deploy this because we won't route traffic to it automatically? 

I'd be interested in getting it reading production traffic but not pushing on to actually ingest so we can run some experiments on how we're chunking and compressing recordings...

hazzadous:
> I came poking about in this repo to figure out how to make exactly this and this PR is sat here like it's already xmas 🎉
>
> I'm guessing it's safe to deploy this because we won't route traffic to it automatically?
>
> I'd be interested in getting it reading production traffic but not pushing on to actually ingest so we can run some experiments on how we're chunking and compressing recordings...

I can't remember what the rollout strategy is but yeah I think something like get the deployment up first with no traffic routed (so we can be sure the ingress isn't pointing to a non-existent deployment on initial deploy), then do another to have things routed.

Please accept this present 🎁  and feel free to get it deployed.

PR #667, opened 2023-01-12
Allow specifying rollout strategy

ellie:
I'd like to test some changes on cloud, and we need to make this configurable for that to happen.

I've set the default values to be the kubernetes defaults, so this should be a no-op until other configuration occurs.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/events-deployment.yaml
+++ b/charts/posthog/templates/events-deployment.yaml
@@ -14,6 +14,13 @@ spec:
   {{- if not .Values.events.hpa.enabled }}
   replicas: {{ .Values.events.replicacount }}
   {{- end }}
+
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: {{ .Values.events.rollout.maxSurge }}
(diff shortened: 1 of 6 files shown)


charts/posthog/templates/events-deployment.yaml
```diff
@@ -14,6 +14,13 @@ spec:
   {{- if not .Values.events.hpa.enabled }}
   replicas: {{ .Values.events.replicacount }}
   {{- end }}
+
```
guidoiaquinti, line 17: Do we want to change events only? What about the others?

ellie: I'm only really thinking of configuring those two atm, but I guess we could add this to everything 🤔 Will add that now

guidoiaquinti: Ah 👍 . I saw the changes in `values.yaml` and I thought some files were missing from the git commit.

ellie: Yeah the plugins deployment is all fancy and templated, so it doesn't need as many file changes

danielxnj (approved):
LGTM

PR #676, opened 2023-02-07
chart: add prometheus-pushgateway for celery metrics

xvello:
## Description

Unlocks https://github.com/PostHog/posthog/pull/14110 by adding the [PushGateway](https://prometheus.io/docs/practices/pushing/) in our prom stack. While this service must be using sparingly, it does fit what we need
to achieve with celery metrics: jobs are scheduled on one of several pods and update one gauge about the region.

We need to only keep the latest value, and the gateway allows us to make that aggregation.

## Type of change

## How has this been tested?

nope

## Checklist

### DIFF
--- a/charts/posthog/Chart.lock
+++ b/charts/posthog/Chart.lock
@@ -32,6 +32,9 @@ dependencies:
 - name: prometheus
   repository: https://prometheus-community.github.io/helm-charts
   version: 15.10.4
+- name: prometheus-pushgateway
+  repository: https://prometheus-community.github.io/helm-charts
+  version: 2.1.1
 - name: prometheus-kafka-exporter
   repository: https://prometheus-community.github.io/helm-charts
   version: 1.6.0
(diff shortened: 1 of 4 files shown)


hazzadous (approved):
Looks good although @guidoiaquinti @ellie is it worth simply adding this via posthog-cloud-infra? I'm not a massive fan of the dev experience with the two updates spanning repos, and triggering a whole new app deploy.

xvello:
Thanks Harry! I just changed the values to not enable it by default, as we don't enable the prom server by default anyway.

Not opinionated about where to install it, but it'll be used for at least one critical alert (the ingestion lag metric), so I think it belongs alonside the rest of the prom stack.

guidoiaquinti (approved):

guidoiaquinti:
> Looks good although @guidoiaquinti @ellie is it worth simply adding this via posthog-cloud-infra? I'm not a massive fan of the dev experience with the two updates spanning repos, and triggering a whole new app deploy.

True, but I think it all depends on how much drift do we want to keep this with cloud.

ellie:
Imo it makes sense to go here, but the dev experience is pretty meh

Would like to revisit deploys _asap_

PR #677, opened 2023-02-07
chart: pass PROM_PUSHGATEWAY_ADDRESS to workers when relevant

xvello:
## Description

Sister PR to https://github.com/PostHog/charts-clickhouse/pull/676, now that the pushgateway is deployed, expose its address to the worker pods, and only them. The pushgateway should be used sparingly, that's why I'm not adding it to common env values.

This assumes that the chart user does not override the service name and port in the `prometheus-pushgateway` subchart values. Alternatively, we could just add to the common envvars in our deploy values. 

## Type of change

## How has this been tested?

Manually added it on dev's worker deploy, and metrics are coming in fine to grafana

## Checklist

### DIFF
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -11,7 +11,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 30.3.5
+version: 30.3.6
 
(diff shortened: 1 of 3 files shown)


hazzadous (approved):
Looks good

I believe this is the way we are handling host name settings for other in Chart dependencies so this seems like a good place as any to do so.

PR #679, opened 2023-02-10
Separate `decide` deployment

ellie:
## Description
Add a new, optional deployment for decide endpoint. If enabled, `/decide` will get its own deployment

*Why?*

Atm it looks like decide can be adversely affected by noisy neighbours. This is not what we want! It also makes debugging much easier if we can view the traffic for this independently

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/ingress.yaml
+++ b/charts/posthog/templates/ingress.yaml
@@ -94,7 +94,7 @@ spec:
             path: "/capture"
             backend: *INGESTION
 
-          {{- if .Values.decide.enabled}}
+          {{- if .Values.decide.ingressEnabled}}
           - pathType: Prefix
             path: "/decide/"
             backend: *INGESTION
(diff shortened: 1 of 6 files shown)


charts/posthog/templates/ingress.yaml
```diff
@@ -94,12 +94,25 @@ spec:
             path: "/capture"
             backend: *INGESTION
 
+          {{- if .Values.decide.enabled}}
```
hazzadous, line 97: When we enable this, will Helm wait for the decide deployment to be ready before updating the ingress?

I have a similar PR [here](https://github.com/PostHog/charts-clickhouse/pull/666) but wasn't sure what the behaviour was.

ellie: ohhh awesome I was going to follow up with this too hahaha

Yeah that is a little worrying. In that case I might adjust this so that we can toggle the ingress separately

danielxnj (approved):

hazzadous (approved):

hazzadous:
🥳

PR #680, opened 2023-02-13
chart: add End2EndIngestionLag and DjangoErrorRate alerts

xvello:
## Description

Part of porting our internal alerts to all regions, https://github.com/PostHog/posthog-cloud-infra/issues/1052. 

- End2EndIngestionLag mirrors `End-to-end ingestion lag exceeds norm`, with the new gauges added last week.
- DjangoErrorRate mirrors `HTTP 500 Response Exceeding avg. 500`, but with a 1% target instead of a 500 errors count. Looking back at prod-us, this would have triggered during the two django incidents we had this month, but not during smaller transient spikes.

For now, the alerts are marked as `warning` to test it, I'll pass it to `critical` and disable the old one in two days, if it does not misbehave.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -1843,7 +1843,7 @@ prometheus:
                 summary: End-to-end analytics event ingestion lag exceeds 5 minutes.
                 description: |
                   Our end-to-end probe measured an ingestion lag higher than 5 minutes for scenario {{ $labels.scenario }}.
-                  Check the Kafka (cluster overview) dashboard to identify what topics and partitions are lagging and
+                  Check the "Kafka (cluster overview)" dashboard to identify what topics and partitions are lagging and
                   follow the https://posthog.com/docs/runbook/services/plugin-server/ingestion-lag runbook for recovery
                   steps.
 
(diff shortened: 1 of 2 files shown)


charts/posthog/values.yaml
```diff
@@ -1819,6 +1819,34 @@ prometheus:
...
+                  identify the issue. Either a dependency is unhealthy, or a recent code change triggered a bug and
+                  needs to be rolled-back.
+
+            - alert: End2EndIngestionLag
+              expr: (max by(scenario) (posthog_celery_observed_ingestion_lag_seconds{scenario=~"ingestion_api|ingestion"})) > 3000
+              for: 2m
+              labels:
+                severity: warning # TODO: move up to critical after testing for two days
+              annotations:
+                summary: End-to-end analytics event ingestion lag exceeds 5 minutes.
```
hazzadous, line 1843: Isn't 3000 seconds 50 minutes rather than 5?

hazzadous (approved):
Looks good aside from the 3000 seconds maybe?

charts/posthog/values.yaml
xvello: Indeed, fixed!

PR #682, opened 2023-02-13
chore(ingestion): add slow lane consumer for bad actors

### LINKED ISSUE #681: Add slow lane consumer for ingestion

hazzadous:
## Description

Sometimes we have _loads_ of events being sent from a single decide,
which plays havoc with us trying to process them in order, and results
in slowing down other users.

We also have cases where people try to import loads of historical
events.

This deployment handles ingestion of these events independently from the
main ingestion consumer thereby allowing for different guarantees on
ingestion latency.

## Type of change

## Checklist
- [x] I have made corresponding changes to the documentation (as in, should be generated from `values.yaml`


charts/posthog/templates/plugins-ingestion-deployment.yaml
```diff
@@ -1,3 +1,18 @@
...
+    The ingestion deployments handle processing of analytic events e.g. captured
+    via the `/e/` and similar endpoints.
+
+    To allow for near real time processing of most of the events, we will divert
+    some events that break certain criteria e.g. sending 1000s of events per
+    second with the same distinct id to a separate consumer to be processed.
+    This is consumer is run by the plugins-ingestion-overflow deployment and can
+    be scaled independently.
+
+    NOTE: before the PostHog app supports the `ingestion-overflow` mode, we'll
```
hazzadous, line 12: Ooops this doesn't work 🤔  the plugin-server will not know what to do for ingestion-overflow

tiina303: If we want to generalise this with async consumer then we can have two params:
1. topic to read
2. pipeline to use

then for 
1. async consumer: it's async topic and async pipeline
2. ingestion: it's ingestion topic and ingestion pipeline (with overflow)
3. ingestion_overflow topic and ingestion pipeline (after overflow step)

charts/posthog/values.yaml
```diff
@@ -474,7 +474,15 @@ pluginsIngestion:
     maxUnavailable: 25%
 
   # -- Additional env variables to inject into the plugin-server stack deployment.
-  env: []
+  env:
+    - name: INGESTION_OVERFLOW_ENABLED
+      # -- Whether or not to send events to a "slow-lane" to be consumed
+      # separately by the pluginsIngestionOverflow deployment. When setting this
+      # to true, you much set pluginsIngestionOverflow.enabled to `true``
```
tomasfarias, line 481: nit: typo

```suggestion
      # to true, you must set pluginsIngestionOverflow.enabled to `true``
```

tomasfarias: Isn't this not a "must" if we enable both `ingestion` and `ingestionOverflow` in the same deployment? Not for us, but thinking about hobby deployments that want to try out overflow (although may be a stretch).

tomasfarias (approved):
Lgtm, deploying this will require https://github.com/PostHog/posthog/pull/14211 to be merged first. We can discuss operational details tomorrow :handshake:

PR #685, opened 2023-02-14
chart: add per-partition lag to kafka dashboard

xvello:
## Description

During ingestion incidents, it's important to quickly determine whether lag is spread across partitions or local.

Adding a second set of graphs with per-partition stats. Empty by default, the user has to select what topic(s) to display, to avoid overwhelming prometheus.

## Type of change

## How has this been tested?

Check it out on [prod-eu with some lag](http://grafana-prod-eu/d/CEEo3dJVz/kafka-cluster-overview-with-partitions?orgId=1&from=1675749129048&to=1675792134266&var-job=kubernetes-pods&var-instance=10.0.135.153:9308&var-topic=All&var-topic_detailed=clickhouse_events_json)

## Checklist

### DIFF
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -11,7 +11,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 30.3.9
+version: 30.4.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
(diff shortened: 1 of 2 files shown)


danielxnj (approved):

hazzadous:
Could this be a separate Dashboard dedicated to ingestion rather than updating the generic Kafka one? I can imagine there will be other more specific details we'd want for the other workloads but could end up making this dashboard unwieldy

hazzadous:
Ah sorry misread it, I thought this was the partition stats James added, discard my comment!

xvello:
Thanks @hazzadous. I've observed several engineers getting into the partition trap these last weeks:

- not catching hot partitions but just looking at the sum
- trying to "fix" a recovery rate going down, while it was due to most partitions catching up.

 That's why I think it's fundamental to call it out in the main dashboard.

hazzadous (approved):

PR #690, opened 2023-02-16
Update decide check to use the correct role

ellie:
## Description

We now have a role just for decide, so use it! Ensures we don't check more than we need

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -11,7 +11,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 30.8.1
+version: 30.8.2
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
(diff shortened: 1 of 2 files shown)


charts/posthog/Chart.yaml
```diff
@@ -11,7 +11,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 30.5.1
+version: 30.5.2
```
hazzadous, line 14: The annoying thing about this is that we are not backwards compatible with older docker images. In this case I _think_ we'll require all dependencies. So would make sense to ensure the upgrade is not rolled out automatically, maybe a minor or major version.

I haven't found a great way to handle this, aside from maybe removing the `role` abstraction and rather allowing specifically the dependencies that should be checked.

I've not got a good answer here 🤷

xvello: That decide pool is not enabled by default, and requires two values to be set before receiving traffic, so I think it's fine to move forward.
We can add a comment to the `decide.enabled` value that it requires posthog 1.44.0 (we'll do this one before switching to continuous releases, right?)

xvello (approved):

charts/posthog/Chart.yaml
hazzadous: Hah yeah forgot it was behind a flag. Yeah adding a note to the release doc sounds like a plan 👍

xvello:
Took the liberty of rebasing and adding the comment about required PH version to the values. Ready to merge!

hazzadous (approved):

PR #693, opened 2023-02-21
feat: Added object storage region

benjackwhite:
## Description
Related to this PR https://github.com/PostHog/posthog-cloud-infra/pull/1145

Adds support for setting the region of the external bucket

@hazzadous quite likely I've done something wrong here so feel free to call it out :D 

## Type of change

## How has this been tested?

Haven't tested and not sure how to 😅 

## Checklist


hazzadous (approved):
Looks good to me.

Have we been falling back to PostgreSQL for this then, and do we need to migrate data from PostgreSQL to S3 now?

benjackwhite:
> Looks good to me.
> 
> Have we been falling back to PostgreSQL for this then, and do we need to migrate data from PostgreSQL to S3 now?

We have been but I'm gonna go ahead and guess not that many exports happened for it to be a problem. Probably worth verifying how much blob data is stored there and if it is a problem

PR #701, opened 2023-02-24
feat(recordings): add separate deployments/hpa for ingestion

hazzadous:
This should allow us to scale session recordings separately from
analytics events ingestion. It is backwards compatible with the existing
deployments setup which doesn't change. Rather we can update chart
values to set:

 1. pluginsIngestion.enabled = false
 2. pluginsAnalyticsIngestion.enabled = true
 3. recordingsIngestion.enabled = true

Which will switch to separate deployments.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -530,6 +530,100 @@ pluginsIngestion:
   sentryDSN:
 
 
+pluginsAnalyticsIngestion:
+  # -- Whether to install the PostHog plugin-server analytics ingestion
+  # capability as an individual workload. Note that this is different from the
+  # `pluginsIngestion` setting above specifically in that the deployment
+  # controlled by these settings does not handle session recordings. This
+  # deployment is intended to be a replacement for `pluginsIngestion` and is
+  # intended to be used along side the `recordingsIngestion` deployment.
+  #
+  # The reason these values were added is to be able to scale analytics and
+  # session recordings events indepentently, and to be able to roll this out in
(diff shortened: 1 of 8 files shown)


charts/posthog/values.yaml
```diff
@@ -530,6 +530,100 @@ pluginsIngestion:
...
+  # -- Additional env variables to inject into the plugin-server stack deployment.
+  env:
+    - name: INGESTION_OVERFLOW_ENABLED
+      # -- Whether or not to send events to a "slow-lane" to be consumed
+      # separately by the pluginsIngestionOverflow deployment. When setting this
+      # to true, you must set pluginsIngestionOverflow.enabled to `true``
+      # otherwise some events will not be processed in cases of overflow. This
+      # behaviour is triggered for instance if there are too many events to be
+      # processed in near real time from a single device.
+      value: "false"
```
tomasfarias, line 579: I thought this was enabled? 🤔

hazzadous: It's enabled in the infra repo [here](https://github.com/PostHog/posthog-cloud-infra/blob/main/helm/values/prod.yml#L230)

hazzadous: Perhaps we should enable be default at some point 🤔  what do you think?

tomasfarias: Ah, okay, I keep getting confused between the two repos. Thank you.

Yeah, I'd say once it has been running for a while we can enable it by default.

PR #706, opened 2023-03-13
chart: add GraphileJobExecutionLag alert

xvello:
## Description

Add a (non-paging for now) monitor on the new `posthog_celery_graphile_lag_seconds` metric.

It links to the runbook draft introduced in https://github.com/PostHog/product-internal/pull/453

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -2221,6 +2221,20 @@ prometheus:
                 summary: Failed to process some session recording events and have been sent to the dead letter queue for review.
                 description: "The `session_recording_events_dlq` topic offset has increased over the past 5 minutes."
 
+            - alert: GraphileJobExecutionLag
+              expr: (max by(task_identifier) (posthog_celery_graphile_lag_seconds{task_identifier!="bufferJob"})) > 300
+              for: 5m
+              labels:
+                rotation: common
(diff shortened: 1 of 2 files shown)


charts/posthog/values.yaml
```diff
@@ -2221,6 +2225,21 @@ prometheus:
                 summary: Failed to process some session recording events and have been sent to the dead letter queue for review.
                 description: "The `session_recording_events_dlq` topic offset has increased over the past 5 minutes."
 
+            - alert: GraphileJobExecutionLag
+              expr: (max by(task_identifier) (posthog_celery_graphile_lag_seconds{task_identifier!="bufferJob"})) > 300
```
tiina303, line 2229: Did we not get rid of the bufferJob ones yet?

tiina303 (approved):
lgtm

charts/posthog/values.yaml
xvello: No, we have not deleted them yet.

tiina303: > and it's better to avoid us paging if we were to send more in the future.

paging yes, slack notification would be nice to have if we started adding these 🤷‍♀️

PR #713, opened 2023-03-23
Add PodDisruptionBudget to decide

frankh:
## Description
Eventually we'll add PDBs to all deploys, but decide shows the most disruption during kubernetes upgrades so we'll start with this one

## Type of change

## Checklist


ellie (approved):
Nice! We should maybe also configure the rollout to scale up a little faster?

frankh:
shouldn't be necessary, during rollouts the rollout configuration takes precedent over the PDB, so it will still be 25% surge and 25% unavailable for deploys

PR #714, opened 2023-03-24
Decide deployment services check

EDsCODE:
**Initial draft changes so discussion can be done here

## Description

based on https://posthog.slack.com/archives/C05034MPYA0/p1679596107221869, the decide pods still rely postgres because of the startup checks. This means that if postgres is down, decide pods will be down and error out regardless of caching

## Discussion points
- Possibly, split up wait checks (postgres // rest of the services) so there's not repeated check scripts for each deployment
- If possible, I think the decide deployment should wait briefly for postgres and migrations and timeout after X amount of time. Not sure the exact mechanics here but we don't want decide deployment to _never_ connect to postgres/pgbouncer. 
- Main point is that postgres isn't _mission critical_ for decide deployment to operate as it can use a populated redis cache so we should not have the deployment be blocked by postgres being up

### DIFF
--- a/charts/posthog/templates/_snippet-initContainers-decide-wait-for-service-dependencies.tpl
+++ b/charts/posthog/templates/_snippet-initContainers-decide-wait-for-service-dependencies.tpl
@@ -1,0 +1,44 @@
+{{/* Common initContainers-decide-wait-for-service-dependencies definition */}}
+{{- define "_snippet-initContainers-decide-wait-for-service-dependencies" }}
+- name: decide-wait-for-service-dependencies
+  image: {{ .Values.busybox.image }}
+  imagePullPolicy: {{ .Values.busybox.pullPolicy }}
+  env:
+    {{- include "snippet.clickhouse-env" . | nindent 4 }}
+  command:
+    - /bin/sh
+    - -c
+    - >
+        {{ if .Values.clickhouse.enabled }}
+        until (
+            NODES_COUNT=$(wget -qO- \
+                "http://$CLICKHOUSE_USER:$CLICKHOUSE_PASSWORD@{{ include "posthog.clickhouse.fullname" . }}.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local:8123" \
+                --post-data "SELECT count() FROM clusterAllReplicas('{{ .Values.clickhouse.cluster }}', system, one)"
+            )
+            test ! -z $NODES_COUNT && test $NODES_COUNT -eq {{ mul .Values.clickhouse.layout.shardsCount .Values.clickhouse.layout.replicasCount }}
+        );
+        do
+            echo "waiting for all ClickHouse nodes to be available"; sleep 1;
+        done
+        {{ end }}
+
+        {{ if .Values.redis.enabled }}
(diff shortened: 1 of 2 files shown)


charts/posthog/templates/_snippet-initContainers-decide-wait-for-service-dependencies.tpl
```diff
@@ -0,0 +1,44 @@
...
+- name: decide-wait-for-service-dependencies
+  image: {{ .Values.busybox.image }}
+  imagePullPolicy: {{ .Values.busybox.pullPolicy }}
+  env:
+    {{- include "snippet.clickhouse-env" . | nindent 4 }}
+  command:
+    - /bin/sh
+    - -c
+    - >
+        {{ if .Values.clickhouse.enabled }}
```
neilkakkar, line 12: Don't need Clickhouse either?

neilkakkar:
> If possible, I think the decide deployment should wait briefly for postgres and migrations and timeout after X amount of time. Not sure the exact mechanics here but we don't want decide deployment to never connect to postgres/pgbouncer.

This just checks the _current_ status of services to determine if the pod can come up, so it doesn't mean we'll never connect to postgres/pgbouncer.

I think instead we'd want something like an "OR" check between redis and postgres? If either is up, we're good to go.

neilkakkar:
> split up wait checks (postgres // rest of the services) so there's not repeated check scripts for each deployment

Good idea, not clear enough about things here, would want @PostHog/team-infra to give their input 😅

tiina303:
Related: looked at the readiness and startup and liveness probes:
1. readiness & startup are both `_readyz?role=decide` ([here](https://github.com/PostHog/charts-clickhouse/blob/main/charts/posthog/templates/decide-deployment.yaml#L155)), which doesn't check PG, but does require cache ([here](https://github.com/PostHog/posthog/blob/bf0552b6d08582a5a091d03105be99a9207437ad/posthog/health.py#L61)), not sure we'd want to (Redis down PG available situations)?
2. liveliness check is [here](https://github.com/PostHog/posthog/blob/bf0552b6d08582a5a091d03105be99a9207437ad/posthog/health.py#L73) - just verifying service is fine 👍

charts/posthog/templates/decide-deployment.yaml
```diff
@@ -170,6 +170,5 @@ spec:
         securityContext: {{- omit .Values.decide.securityContext "enabled" | toYaml | nindent 12 }}
         {{- end }}
       initContainers:
-      {{- include "_snippet-initContainers-wait-for-service-dependencies" . | indent 8 }}
-      {{- include "_snippet-initContainers-wait-for-migrations" . | indent 8 }}
```
hazzadous, line 174: Decide still makes queries to PostgreSQL right? So we still need to wait for them to have been applied?

tiina303: I think I might have misunderstood earlier & mislead the ff team ... this would only apply on new container setup, not on restart? If that's true and considering we do rolling deploys this is fine to keep as is.

hazzadous: Yeah if you don't include the migration check, you'll end up with a `SELECT all, the, fields, including, ones, that, don't, exist, yet FROM posthog_persons` which will error and if we're lucky fall back to redis, or just straight error out. Either way I don't think that's desired.

frankh: just confirmed that init containers only run when the pod is created

pods never restart, only the containers within them

PR #717, opened 2023-04-05
Add pgbouncer read deployment

ellie:
This should be pointed to a read replica for a postgres cluster

1. Why not RDS proxy? A few reasons. Turns out, it's super expensive! Classic AWS. Otherwise, it's pretty opaque + things like decide really need to know how the pooling works
2. Why do we need another deployment? pgbouncer doesn't really have query routing :( There is a fork called [pgbouncer-rr](https://github.com/awslabs/pgbouncer-rr-patch) that does what we need, but I'm not that keen on using a fork. Plus, two deployments can reduce the risk of things going wrong and _totally_ bringing us down.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -1217,7 +1217,8 @@ externalPostgresql:
 ### ---- PGBOUNCER ----
 ###
 ###
-pgbouncer:
+
+_pgbouncer: &_pgbouncer 
   # -- Whether to deploy a PgBouncer service to satisfy the applications requirements.
   enabled: true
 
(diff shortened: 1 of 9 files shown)


charts/posthog/values.yaml
```diff
@@ -1337,6 +1338,12 @@ pgbouncer:
   ## -- PgBouncer pod(s) annotation.
   podAnnotations: {}
 
+pgbouncer:
+  <<: *_pgbouncer
+
+pgbouncer_read:
```
hazzadous, line 1344: Maybe for consistency:

```suggestion
pgbouncerRead:
```

Although I don't know what the generally used format is in the community.

ellie: Ahh good point, will change it

ellie:
@hazzadous if you could double 👀 the host override stuff please that would be awesome! I'd like to enable in dev first and see what happens

the rest of my team are off so no reviews from them :(

PR #718, opened 2023-04-05
feat(keda): add ability to scale recordings ingestion with KEDA

hazzadous:
This just adds the Kafka triggers. I've left it pretty free form on what
you can pass in to the spec as doesn't seem much benefit in abstracting
that but it would be annoying to need to explicitly add functionality
that we might want to use in a deployment.

## Type of change

## Checklist


fuziontech (commented):
We have KEDA now?! 🥳

PR #719, opened 2023-04-06
fix(recordings): fix types using in KEDA ScaledObject

hazzadous:
Previously I'd used booleans instead of strings as it expects. It was
also expecting a comma separated list for the brokers list. We were
being inconsistent here between chart internal an external Kafka
settings.

## Type of change

## Checklist


charts/posthog/templates/recordings-ingestion-deployment.yaml
```diff
@@ -29,7 +29,7 @@ spec:
       bootstrapServers: {{ include "posthog.kafka.brokers" . }}
       consumerGroup: session-recordings  # TODO: pass this into the consumers as well to avoid them getting out of sync.
       topic: session_recording_events  # TODO: pass this into the consumers as well to avoid them getting out of sync.
-      allowIdleConsumers: false  # Don't scale past the number of partitions
+      allowIdleConsumers: "false"  # Don't scale past the number of partitions
```
ellie, line 32: but why 😭

PR #720, opened 2023-04-13
feat: Add temporal and temporal worker to helm stack

fuziontech:
This is very much a WIP - avert your eyes!

- feat: add temporal worker to helm
- checkpoint

## Description
Add a Temporal scheduler based on Temporal's helm chart
https://github.com/temporalio/helm-charts

In production we will be using Temporal Cloud's scheduler, but it's important to include Temporal scheduler in the helm chart for the Open Source users out there (and for testing!)

This also adds a python Temporal worker that will be used to execute workflows on our k8s stack.

## How has this been tested?
HAH - we test in prod.

Not really - there will be tests and this will have been tested extensively on DO both with local Temporal scheduler and with Temporal.cloud

### DIFF
--- a/charts/posthog/templates/temporal-py-worker-hpa.yaml
+++ b/charts/posthog/templates/temporal-py-worker-hpa.yaml
@@ -1,0 +1,25 @@
+{{- if and .Values.temporalPyWorker.enabled .Values.temporalPyWorker.hpa.enabled -}}
+apiVersion: autoscaling/v2beta2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ template "posthog.fullname" . }}-worker
+  labels: {{- include "_snippet-metadata-labels-common" . | nindent 4 }}
+spec:
+  scaleTargetRef:
+    kind: Deployment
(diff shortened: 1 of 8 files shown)


ellie (approved):

charts/posthog/templates/temporal-py-worker-hpa.yaml
```diff
@@ -0,0 +1,25 @@
...
+apiVersion: autoscaling/v2beta2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ template "posthog.fullname" . }}-worker
+  labels: {{- include "_snippet-metadata-labels-common" . | nindent 4 }}
+spec:
+  scaleTargetRef:
+    kind: Deployment
+    apiVersion: apps/v1
+    name: {{ template "posthog.fullname" . }}-temporalPyWorker
```
ellie, line 11:
```suggestion
    name: {{ template "posthog.fullname" . }}-temporal-py-worker
```

if what I read above is correct

fuziontech:
very old

PR #726, opened 2023-04-19
feat: Added recording blob ingestion deployment

benjackwhite:
## Description

Adds a new deployment for the recording blob ingestion

## Type of change

## Checklist


charts/posthog/values.yaml
```diff
@@ -746,6 +746,108 @@ recordingsIngestion:
...
+    # config such as consumer group id and topic which will be set
+    # automatically.
+    #
+    # TODO: add support for all security protocol config.
+    kafkaTrigger:
+      enabled: true
+
+      metadata:
+        # -- At what lag value should KEDA scaling start.
+        lagThreshold: "1000"
```
hazzadous, line 793: These defaults will probably be slightly different but we can play with that later.

charts/posthog/templates/recordings-blob-ingestion-deployment.yaml
```diff
@@ -0,0 +1,38 @@
+{{ if .Values.recordingsBlobIngestion.enabled }}
```
hazzadous, line 1: Can you add a unittest for this deployment? Just basic tests. I guess specific ones for S3 details would be sensible.

hazzadous: Yeah pretty much 😬  but perhaps tailor to the specific env vars you abs need.

It will at least ensure that enable: true etc works as expected.

benjackwhite: I added a test but not for the envs, largely because they aren't different to the existing object storage ones.

charts/posthog/values.yaml
benjackwhite: Yeah true

hazzadous (approved):

PR #729, opened 2023-04-25
feat(chart): add new externalKafka.tls value

xvello:
## Description

To run KEDA autoscaling on Kafka lag, we will need to set it to use TLS when needed. We currently set the `KAFKA_SECURITY_PROTOCOL` envvar in common values, but retrieving that value to setup KEDA is non-trivial. Instead:

- add a new `externalKafka.tls` value
- if it as set, and we don't use the chart's brokers, set the `KAFKA_SECURITY_PROTOCOL` envvar to `SSL`
- use the same value to setup `tls: enabled` on the KEDA Kafka trigger
- also set the KEDA Kafka trigger to use an absolute `Value` metric instead of the default `AverageValue` (we want to set an absolute acceptable lag value for the topic, not a lag per container)

This is expected to be a backwards-compatible change. Once this is released, we can change our values to not set the envvar, and set the new value instead.

## Type of change

## How has this been tested?

Added a unit test case, and updated existing ones

## Checklist

### DIFF
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -11,7 +11,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 30.14.2
+version: 30.15.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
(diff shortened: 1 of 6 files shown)


ellie (approved):
LGTM!

hazzadous (approved):

PR #731, opened 2023-04-27
feat: Added volume mounts to chart for blob data

benjackwhite:
## Description

Adds volume mount configurations for recording blob ingestion

## Type of change

## How has this been tested?
With tests 👍 

## Checklist


frankh (approved):

charts/posthog/templates/_snippet-plugins-deployment.tpl
```diff
@@ -96,6 +100,11 @@ spec:
         # Expose the port on which the healtchheck endpoint listens
         - containerPort: 6738
         env:
+        {{ if .params.volumeMounts }}
```
hazzadous, line 103: So much abstraction/templating! I was the person that added the templating here, I feel like eventually we should just have separate definitions, but that’s ot

hazzadous (approved):

PR #733, opened 2023-04-28
feat(capture): disable init container for event deployment

xvello:
## Description

Remove both the `wait-for-migrations` and `wait-for-service-dependencies` init containers for the events deployment.

We want event pods to be able to upscale even if the whole platform is on fire (including PG, CH and redis). Their only dependency is Kafka, and it is asserted in the readiness probe.

A bad deploy that kills the Kafka client should not go through, because the readiness probe will fail, triggering a rollback.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/events-deployment.yaml
+++ b/charts/posthog/templates/events-deployment.yaml
@@ -101,6 +101,12 @@ spec:
         # statsd env variables
         {{- include "snippet.statsd-env" . | indent 8 }}
 
+        # async migrations will be run by the migrate job
+        {{- if .Values.migrate.enabled -}}
+        - name: SKIP_ASYNC_MIGRATIONS_SETUP
+          value: 'True'
+        {{- end }}
+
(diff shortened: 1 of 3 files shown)


tiina303 (approved):
🚢

hazzadous (approved):
Looks good although you have a test failure.

charts/posthog/templates/events-deployment.yaml
```diff
@@ -101,6 +101,12 @@ spec:
         # statsd env variables
         {{- include "snippet.statsd-env" . | indent 8 }}
 
+        # async migrations will be run by the migrate job
+        {{- if .Values.migrate.enabled -}}
+        - name: SKIP_ASYNC_MIGRATIONS_SETUP
+          value: 'True'
+        {{- end }}
```
tiina303, line 108: might cause some problems if self-hosted users try to bypass too far to the future, but that's low priority, so 🚢  🚀

PR #734, opened 2023-05-03
fix: Correct podSecurityContext for blob ingester

benjackwhite:
## Description

We realised that PostHog's container runs in a non-root group (1000) so we need to set this default for the blob ingster.

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -832,7 +832,8 @@ recordingsBlobIngestion:
     enabled: false
   # -- Pod security context for the plugin-server stack deployment.
   podSecurityContext:
-    enabled: false
+    enabled: true
+    fsGroup: 1000
 
   livenessProbe:
     # -- The liveness probe failure threshold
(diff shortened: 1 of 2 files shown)


charts/posthog/values.yaml
```diff
@@ -832,7 +832,8 @@ recordingsBlobIngestion:
     enabled: false
   # -- Pod security context for the plugin-server stack deployment.
   podSecurityContext:
-    enabled: false
+    enabled: true
```
frankh, line 835: is this intentional? that will turn it on for every deploy

benjackwhite: Is it not intentional? As far as I understand we need the security context for the volume access so without this it wouldn't work?

pauldambra: But only for every deploy of the blob ingester?

frankh: sorry i wasn't paying attention i thought you were setting the top level enabled to on 🤦

PR #742, opened 2023-05-26
chore: don't resolve jobs backpressure page if data is missing

fuziontech:
## Description
We are getting paged on repeat when graphile job queue is falling behind because Prom resets the alert on no data. This should keep that from happening.

## Type of change

## Checklist


charts/posthog/values.yaml
```diff
@@ -2426,7 +2426,7 @@ prometheus:
                 description: "The `session_recording_events_dlq` topic offset has increased over the past 5 minutes."
 
             - alert: GraphileJobExecutionLag
-              expr: (max by(task_identifier) (posthog_celery_graphile_lag_seconds{task_identifier!="bufferJob"})) > 900
```
xvello, line 2429: Heya, we still want the `by(task_identifier)`, otherwise a second type lagging will not trigger another alert

xvello (approved):

PR #747, opened 2023-06-16
very very WIP

pauldambra:
We want to add separate session recording kafka config so that we can move session recordings workloads to a new cluster


pauldambra:
a bunch of the kube tests are failing... I guess because https://github.com/PostHog/posthog/pull/16109 is necessary before this change makes sense

benjackwhite (approved):
Looks great! Pretty sure the tests will pass once we have the corresponding Posthog PR in

pauldambra:
@hazzadous this is the second step in our episode of "having fun with kafka"

I think this sets us up so we can start to set a separate session recording kafka cluster

frankh (approved):

PR #748, opened 2023-06-20
feat: Add support for external replay redis

benjackwhite:
## Description

For cloud we want to use a different external redis just for session recording. Kept it simple and added only host / port support

## Type of change

## How has this been tested?
Tests

## Checklist


charts/posthog/templates/_snippet-plugins-deployment.tpl
```diff
@@ -122,6 +122,9 @@ spec:
         # Redis env variables
         {{- include "snippet.redis-env" .root | nindent 8 }}
 
+        # Session Recoridng Redis env variables
```
pauldambra, line 125:
```suggestion
        # Session Recording Redis env variables
```

pauldambra (approved):
💪

PR #753, opened 2023-06-23
add new recordings role for django

xvello:
## Description

We want to isolate failures between `/s` and `/e`, by running them in separate resource pools.

This PR:
- copies the specs and tests of the `events` deploy into a `recordings` deploy.
- adds the `recordings.ingressEnabled` value to route `/s` to the new deploy, mirroring how we implemented the decide split
- updates the kubetest values to avoid flakes caused by resource limits

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/templates/recordings-deployment.yaml
+++ b/charts/posthog/templates/recordings-deployment.yaml
@@ -1,0 +1,178 @@
+{{- if .Values.recordings.enabled -}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "posthog.fullname" . }}-recordings
+  labels: {{- include "_snippet-metadata-labels-common" . | nindent 4 }}
+  annotations: {{- include "_snippet-metadata-annotations-common" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+        app: {{ template "posthog.fullname" . }}
+        release: "{{ .Release.Name }}"
+        role: recordings
+  {{- if not .Values.recordings.hpa.enabled }}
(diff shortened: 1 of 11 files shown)


hazzadous (approved):

hazzadous (approved):
You'll need to update the chart version though...

hazzadous (approved):

charts/posthog/templates/recordings-deployment.yaml
```diff
@@ -0,0 +1,178 @@
...
+          initialDelaySeconds: {{ .Values.web.readinessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.web.readinessProbe.periodSeconds }}
+          successThreshold: {{ .Values.web.readinessProbe.successThreshold }}
+          timeoutSeconds: {{ .Values.web.readinessProbe.timeoutSeconds }}
+        startupProbe:
+          httpGet:
+            # Now that capture's only external dependency is Kafka, let's use the
+            # reduced readiness check as its startupProbe. This means that we can
+            # deploy new configs / versions even when Postgres and Redis are down.
+            path: /_readyz?role=events
```
pauldambra, line 158: this makes me wonder if we should have a specific readiness check for recordings events here 🤔 

(not blocking for this, just my ever growing to-do list 🤣 )

pauldambra (approved):

charts/posthog/templates/recordings-deployment.yaml
xvello: It's indeed something to keep in mind. `role=events` is only checking for kafka now, but if you're writing to a different MSK, you'll need to check that one too.

PR #755, opened 2023-06-27
fix(plugin-server): use /_health for liveness

hazzadous:
We were just using exec and `true` command before, but maybe this will
kill pods if they stop consuming on e.g. `exports`.

Has the potential to cause more issues than it solves without some 
tuning, but we could give it a go.

@danielxnj where do I make changes in the other chart?

## Type of change

## Checklist


fuziontech (approved):
_yesssss_

PR #775, opened 2023-10-18
Don't wait for migrations for capture pods

frankh:
## Description
These pods don't care about migrations and waiting for them slows down scaling and deployments

## Type of change

## Checklist

### DIFF
--- a/charts/posthog/Chart.yaml
+++ b/charts/posthog/Chart.yaml
@@ -11,7 +11,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 30.31.0
+version: 30.32.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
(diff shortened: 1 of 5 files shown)


xvello (approved):

charts/posthog/values.yaml
```diff
@@ -16,7 +16,7 @@ image:
   # -- PostHog image tag to use (example: `release-1.43.0`).
   tag:
   # -- PostHog default image. Do not overwrite, use `image.sha` or `image.tag` instead.
-  default: ":release-1.43.0"
+  default: ":latest"
```
xvello, line 19: we have stopped pushing `lastest`, let's use `master` here?

xvello (approved):

charts/posthog/values.yaml
frankh: this is posthog/posthog, where we push latest but not master by the looks of it https://hub.docker.com/r/posthog/posthog/tags?page=1&name=master

PR #780, opened 2023-12-06
feat: Make Celery workers and queues configurable

webjunkie:
## Description

We want to run separate Celery workers that consume a different set of queues each.
See RFC https://github.com/PostHog/product-internal/pull/537

## Type of change
- New feature (non-breaking change which adds functionality)

## Changes
- add possibility to configure several workers, then use range in the template
  - please advise if this is the way to go or not at all
- we can pass the config options as environment variables, which override Celery default settings
- by default, nothing should change right now
  - but allows that we can put a config into dev environment and then see how it works

Example config for dev later:

```yaml
  workers:
    - name: priority
      queues:
        - email
    - name: default
      queues:
        - celery # default queue for Celery
        - insight_refresh
        - insight_export
    - name: gevent
      pool: gevent
      concurrency: 50
      queues:
        - analytics_queries
```

## How has this been tested?

- how can I properly test this?

## Checklist

### DIFF
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -301,7 +301,7 @@ worker:
   # -- Count of worker pods to run. This setting is ignored if `worker.hpa.enabled` is set to `true`.
   replicacount: 1
 
-  workers:
+  consumers:
     - name: default
 
   hpa:
--- a/charts/posthog/values.yaml
+++ b/charts/posthog/values.yaml
@@ -301,6 +301,22 @@ worker:
   # -- Count of worker pods to run. This setting is ignored if `worker.hpa.enabled` is set to `true`.
   replicacount: 1
 
+  workers:
+    - name: priority
+      queues:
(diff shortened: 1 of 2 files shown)


charts/posthog/values.yaml
```diff
@@ -301,6 +301,9 @@ worker:
   # -- Count of worker pods to run. This setting is ignored if `worker.hpa.enabled` is set to `true`.
   replicacount: 1
 
+  workers:
```
benjackwhite, line 304: Perhaps we want "consumers" just to avoid the multipe use of the word "workers"?

`Our "worker" chart deploys "workers" each of which can have multiple celery "workers"` 😅

webjunkie: That's better, yes!

webjunkie:
@frankh Can you take a look here?

frankh (approved):

